PR #3, opened 2020-09-07
Implement artifacts

woksin:
Expands upon Artifacts.

I had to make the Artifact class not abstract to make it like we have in the JS SDK.


Source/Artifacts/Artifact.cs
```diff
@@ -9,8 +9,18 @@ namespace Dolittle.SDK.Artifacts
     /// <summary>
     /// Represents the concept of an artifact.
     /// </summary>
-    public abstract class Artifact : Value<Artifact>
+    public class Artifact : Value<Artifact>
```
jakhog: Not sure I understand why it needs to be !abstract. We shouldn't really ever be instantiating it, should we?

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,119 @@
...
+        public void Associate(Type type, Artifact artifact)
+            => _registered.OnNext(new ArtifactAssociation(type, artifact));
+
+        /// <inheritdoc />
+        public void Associate(Type type, ArtifactId artifactId)
+            => Associate(type, new Artifact(artifactId, Generation.First));
+
+        /// <inheritdoc />
+        public void Associate(Type type, ArtifactId artifactId, Generation generation)
+            => Associate(type, new Artifact(artifactId, generation));
```
jakhog: Do we really need these methods? This seems like some conversions that should be done elsewhere. This isn't really a thing that developers will see, right?

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <inheritdoc />
+        public Type GetTypeFor(Artifact artifact)
+        {
+            var type = _associations.Value.FirstOrDefault(_ => _.Value == artifact).Key;
+            if (type == default) throw new UnknownType(artifact);
+            return type;
+        }
+
+        /// <inheritdoc />
+        public Type GetTypeFor(ArtifactId artifactId) => GetTypeFor(new Artifact(artifactId));
```
jakhog: I think this method should go - as it doesn't really make sense to get a type for just an _id_. It doesn't have a type, only when you specify that it is the first generation - which feels very hidden here.

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,119 @@
...
+        public bool HasFor<T>() => HasFor(typeof(T));
+
+        /// <inheritdoc />
+        public bool HasFor(Type type) => _associations.Value.ContainsKey(type);
+
+        /// <inheritdoc />
+        public bool HasTypeFor(Artifact artifact) => _associations.Value.Any(_ => _.Value == artifact);
+
+        /// <inheritdoc />
+        public bool HasTypeFor(ArtifactId artifactId) => HasTypeFor(new Artifact(artifactId));
```
jakhog: Same here as above - don't think this should be possible.

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,119 @@
...
+            _associations = new BehaviorSubject<IDictionary<Type, Artifact>>(new Dictionary<Type, Artifact>());
+            _logger = logger;
+
+            foreach ((var type, var artifact) in associations)
+            {
+                _registered.OnNext(new ArtifactAssociation(type, artifact));
+            }
+
+            _registered.Subscribe(AddAssociation);
+        }
```
jakhog: Maybe cleaner to call Associate from the outside (the builder) and just have the subscription set up here?

jakhog: Also - no need to explicitly use the generic logger - that's just for creating new instances. This thing doesn't really care about that.

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Dolittle.SDK.Artifacts
+{
+    /// <summary>
+    /// Represents an implementation of <see cref="IArtifacts" />.
+    /// </summary>
+    public class Artifacts : IArtifacts, IDisposable
+    {
+        readonly ReplaySubject<ArtifactAssociation> _registered;
+        readonly BehaviorSubject<IDictionary<Type, Artifact>> _associations;
```
jakhog: Seems a bit superflous to keep track of all the registered artifacts twice? Escpecially if none of them are exposed to the outside world.

We could change the _registered to just be a normal subject, and the _associations to be a BehaviourSubject (which we could allow to outside code to subscribe to for updates.

Source/Artifacts/TypeCannotHaveMultipleArtifactAssociations.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.SDK.Artifacts
+{
+    /// <summary>
+    /// Exception that gets thrown when a <see cref="Type" /> is associated with multiple <see cref="Artifact" />.
+    /// </summary>
+    public class TypeCannotHaveMultipleArtifactAssociations : Exception
```
jakhog: Should there also be a converse error - so an Artifact cannot be associated to multiple types?
It seems like that is just ignored right now.

jakhog (changes requested):

Source/Artifacts/Artifact.cs
woksin: If we do it like we do in the JS SDK this needs to be !abstract. Unless we only have methods that takes in Artifact and not ArtifactId. It makes more sense to have it !abstract if we go the route where we extend the Artifact class with EventType, etc...

We can go the route of having it abstract, but then we need to have an EventType class extending Artifact and have a builder thing for EventType that can take in an ArtifactId and register it to the Artifacts

Source/Artifacts/Artifacts.cs
woksin: Kinda related to above comment. No we don't need the functionality here, but we should be able to register event types (artifacts) by just passing in a guid string.

Source/Artifacts/Artifacts.cs
woksin: Agree. I just mirrored the JS SDK

Source/Artifacts/Artifacts.cs
woksin: Same as above comment

Source/Artifacts/Artifacts.cs
woksin: Isn't it best for a class to own a ILogger for its own Type? To make sure that the logger it uses doesn't print anything weird, for instance using another class (category)

woksin: > Maybe cleaner to call Associate from the outside (the builder) and just have the subscription set up here?

Sure we can do that. Again just mirrored the JS SDK implementation

Source/Artifacts/Artifacts.cs
woksin: That we can do. Makes sense

Source/Artifacts/Artifacts.cs
jakhog: Yes, but I think that is just as well served with builder methods to handle these cases. And in any case, you don't register artifacts, but event types (the only one right now).

Source/Artifacts/TypeCannotHaveMultipleArtifactAssociations.cs
woksin: It is ignored, also ignored in JS SDK, We should track it IMO and throw an exception.

Should we check for multiple associations in the Artifacts implementaiton as well?

Source/Artifacts/Artifacts.cs
woksin: True. The builder can be more flexible. But still does not answer the question of whether the Artifact should be abstract or not

woksin: I do though agree that it should be abstract. Let me expand it with EventType and see where that goes :)

Source/Artifacts/IArtifacts.cs
```diff
@@ -10,6 +12,11 @@ namespace Dolittle.SDK.Artifacts
     /// </summary>
     public interface IArtifacts
     {
+        /// <summary>
+        /// Gets the <see cref="BehaviorSubject{T}" /> of associated artifacts.
+        /// </summary>
+        BehaviorSubject<IDictionary<Type, Artifact>> Associations { get; }
```
jakhog: ...Subject implementations should not be exposed to the outside - they are just useful implementations of both IObservable and IObserver with usefull behaviours. Exposing the ...Subject allows outside things to emit values out from the subject as well, which is usually not what you want.
```suggestion
        IObservable<IDictionary<Type, Artifact>> Associations { get; }
```

Source/Artifacts/Artifacts.cs
```diff
@@ -14,74 +14,69 @@ namespace Dolittle.SDK.Artifacts
...
+            _registered = new Subject<ArtifactAssociation>();
+            Associations = new BehaviorSubject<IDictionary<Type, Artifact>>(new Dictionary<Type, Artifact>());
             _logger = logger;
 
-            foreach ((var type, var artifact) in associations)
-            {
-                _registered.OnNext(new ArtifactAssociation(type, artifact));
-            }
-
             _registered.Subscribe(AddAssociation);
```
jakhog: If you want to do this logic in a slightly more _reactive canonical_ way, you can use the `Scan` operator. So without trying out any code, something like this:
```
_registered.Scan((map, association) =>
  {
    map.Add(association.Type, association.Artifact);
    return map;
  }, new Dictionary<...>()).Subscribe(Associations);
```

jakhog (changes requested):
Ahh, now I think I see why having EventType as a class is making things more complicated.

I'm starting to wonder if that is not the right way to go after all, it seems to just be causing more trouble after all.

Maybe it is better to go back to use Artifact as the type, but use EventType in builder method names, and in the properties, so for events I guess it will also show up in the EventContext. I'm not really sure what I feel here. But I see there is quite a lot more involved than I originally thought of here...

woksin:
Yes, and as @joelhoisko also discovered with serializing artifacts

Source/Artifacts/Artifacts.cs
woksin: Cooool

Source/Artifacts/IArtifacts.cs
woksin: I thought that exposing the IObserver part could be cook. But maybe that adds too much complexity

woksin:
While it makes things a little bit harder I think that it's cool. It's still clear and easy for the user (and when logging the Artifact) it says it's EventType.

But were you thinking of just removing abstract and keeping the EventType : Artifact?

Source/Artifacts/Artifacts.cs
woksin: That signature is wrong. The callback there is Func<ArtifactAssociation, ArtifactAssociation, ArtifactAssociation>

woksin:
I don't know what I think of this abstract Artifact thing. It may also be that the way I do it now is weird. Please have a look @einari @joelhoisko @jakhog

Source/Artifacts/Artifacts.cs
jakhog: Small typo from my side, the new Dictionary thing should have been before the callback (seed value). Then it works.

jakhog:
At this point in time - I am confusing myself, and thinking too much. Maybe we should just pull this in and see how it impacts the rest of the code? If it becomes a pain, it's not a big deal to go back on it...

woksin:
I think that the pain is related to the Artifact being abstract. It doesn't need to be. Whether it makes sense depends on the scenario. In a pure SDK scenario it makes sense to be abstract, while in a protobuf scenario it might make sense to have it non-abstract.

As you said we should try it out. I'll try to use the Scan method again first though

Source/SDK/Client.cs
```diff
@@ -32,6 +36,11 @@ public class Client
         /// </summary>
         public IExecutionContextManager ExecutionContextManager { get; }
 
+        /// <summary>
+        /// Gets the <see cref="EventTypes" />.
+        /// </summary>
+        public EventTypes Artifacts { get; }
```
joelhoisko:
```suggestion
        /// <summary>
        /// Gets the <see cref="EventTypes" />.
        /// </summary>
        public EventTypes EventTypes { get; }
```

This property should probably get renamed to EventTypes.

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,81 @@
...
+        public bool HasFor(Type type) => _typeToArtifactMap.ContainsKey(type);
+
+        /// <inheritdoc />
+        public bool HasTypeFor(TArtifact artifact) => _artifactToTypeMap.ContainsKey(artifact);
+
+        void AddAssociation(Type type, TArtifact artifact)
+        {
+            ThrowIfMultipleTypesAssociatedWithArtifact(artifact);
+            ThrowIfMultipleArtifactsAssociatedWithType(type);
+            _logger.LogDebug("Associating {Type} to {Artifact}", type, artifact);
```
joelhoisko:
```suggestion
        void AddAssociation(Type type, TArtifact artifact)
        {
            _logger.LogTrace("Associating {Type} to {Artifact}", type, artifact);
            ThrowIfMultipleTypesAssociatedWithArtifact(artifact);
            ThrowIfMultipleArtifactsAssociatedWithType(type);
```
This should be a trace level message.

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        void ThrowIfMultipleTypesAssociatedWithArtifact(TArtifact artifact)
+        {
+            if (_artifactToTypeMap.ContainsKey(artifact)) throw new CannotHaveMultipleTypesAssociatedWithArtifact(artifact);
+        }
+
+        void ThrowIfMultipleArtifactsAssociatedWithType(Type type)
+        {
+            if (_typeToArtifactMap.ContainsKey(type)) throw new CannotHaveMultipleArtifactsAssociatedWithType(type);
+        }
```
joelhoisko: We should pass the already registered type/artifact to the exception so that it's easier to find it.

Source/Artifacts/Artifacts.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Artifacts{TArtifact}"/> class.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger" />.</param>
+        protected Artifacts(ILogger<Artifacts<TArtifact>> logger)
+        {
+            _logger = logger;
+            _artifactToTypeMap = new Dictionary<TArtifact, Type>();
+            _typeToArtifactMap = new Dictionary<Type, TArtifact>();
+        }
```
joelhoisko:
```suggestion
    public abstract class Artifacts<TArtifact> : IArtifacts<TArtifact>
        where TArtifact : Artifact
    {
        readonly IDictionary<TArtifact, Type> _artifactToTypeMap;
        readonly IDictionary<Type, TArtifact> _typeToArtifactMap;
        readonly ILogger _logger;

        /// <summary>
        /// Initializes a new instance of the <see cref="Artifacts{TArtifact}"/> class.
        /// </summary>
        /// <param name="logger">The <see cref="ILogger" />.</param>
        protected Artifacts(ILogger logger)
        {
            _logger = logger;
            _artifactToTypeMap = new Dictionary<TArtifact, Type>();
            _typeToArtifactMap = new Dictionary<Type, TArtifact>();
        }
```
We don't need to specify the type of the given `ILogger`, the class doesn't really care what's the type of the logger, just that it's created for it. It is still given the correct typed logger from the factory.

Source/Artifacts/Artifacts.csproj
```diff
@@ -4,6 +4,11 @@
   <PropertyGroup>
     <AssemblyName>Dolittle.SDK.Artifacts</AssemblyName>
   </PropertyGroup>
+    
+  <ItemGroup>
+    <PackageReference Include="System.Reactive" Version="4.4.1" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
+  </ItemGroup>
```
joelhoisko:
```suggestion
  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
  </ItemGroup>
```
Reactive not used anymore and we can just use the abstractions logging.

Source/Artifacts/IArtifacts.cs
```diff
@@ -6,86 +6,59 @@
...
         /// </summary>
         /// <param name="type">CLR type of the artifact.</param>
         /// <returns><see cref="Artifact"/>.</returns>
-        Artifact GetFor(Type type);
+        TArtifact GetFor(Type type);
 
         /// <summary>
         /// Get a CLR <see cref="Type"/> for a specific <see cref="Artifact"/>.
         /// </summary>
         /// <param name="artifact"><see cref="Artifact"/> to get for.</param>
```
joelhoisko:
```suggestion
        /// Check if there is an <typeparamref name="TArtifact"/> associated with a <see cref="Type" />.
        /// </summary>
        /// <param name="artifact">The <typeparamref name="TArtifact"/>.</param>
```

All of the other references to `TArtifact` in the comments should also use `<typeparamref name="TArtifact"/>`

Source/Events/EventType.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Dolittle.SDK.Artifacts;
+
+namespace Dolittle.SDK.Events
+{
+    /// <summary>
+    /// Represens the an event type <see cref="Artifact" />.
```
joelhoisko:
```suggestion
    /// Represents the type of an event.
```

Source/Events/EventTypes.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// Represents a system that knows about<see cref="EventType" />.
+    /// </summary>
+    public class EventTypes : Artifacts<EventType>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EventTypes"/> class.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger" />.</param>
+        public EventTypes(ILogger<Artifacts<EventType>> logger)
```
joelhoisko:
```suggestion
        public EventTypes(ILogger logger)
```
Don't need to specify the logger types here either.

Source/Events/Events.csproj
```diff
@@ -3,7 +3,11 @@
     <PropertyGroup>
         <AssemblyName>Dolittle.SDK.Events</AssemblyName>
     </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
```
joelhoisko:
```suggestion
        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
```

joelhoisko (changes requested):
Looks good, some comments regarding documentation, logging etc

Source/Artifacts/Artifacts.cs
woksin: Solved by https://github.com/dolittle/DotNET.SDK/pull/3/commits/894222a267acc5260eaf684c05a70db420a20a61

woksin:
The last four commits here changes things up again. Could say it is a drawback of being too generic with having the Artifact structures, but I kinda think this is cool. I also think that I got this nicely converting between artifacts and protobuf in my next PR #6

joelhoisko (approved):

PR #4, opened 2020-09-08
Add Rules, Collections, Transactions and Reflection to the SDK

joelhoisko:
Adds  everything from Rules and Collections from Fundamentals under `Dolittle.SDK.Rules`, `Dolittle.SDK.Collections`, `Dolittle.SDK.Transactions`, `Dolittle.SDK.Reflection` packages.


woksin:
I don't know if we really need all of this? Some of it is nice to have, like the NullFree collections and stuff.

Do we need the Rules for anything in the SDK?

jakhog:
I don't think we really need any of this? Did you need it for the EventStore?

joelhoisko:
Not needed.

PR #5, opened 2020-09-08
Implement ReverseCallClient with ReactiveX

jakhog:
Added ReverseCallClient implementation, and copied some Protobuf converters that was needed.

I have not tried the implementation - or made any specs. But the interfaces should be ready for use elsewhere.


Source/Services/IAmADuplexStreamingMethod.cs
```diff
@@ -0,0 +1,35 @@
...
+        where TClient : ClientBase<TClient>
+        where TClientMessage : IMessage
+        where TServerMessage : IMessage
+    {
+        /// <summary>
+        /// Initializes a new instance of the <typeparamref name="TClient"/> client given a <see cref="Channel"/>.
+        /// </summary>
+        /// <param name="channel">The <see cref="Channel"/> to use to connect to the server.</param>
+        /// <returns>A <typeparamref name="TClient"/> client that can be used to perform the call.</returns>
+        TClient Client(Channel channel);
```
woksin: Maybe a name that's more familiar? Like ClientFrom? CreateClient?

Source/Services/IAmADuplexStreamingMethod.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Performs the duplex streaming method call on the server.
+        /// </summary>
+        /// <param name="client">The <typeparamref name="TClient"/> client to use to perform the call.</param>
+        /// <param name="callOptions">The <see cref="CallOptions"/> to use for the call.</param>
+        /// <returns>The <see cref="AsyncDuplexStreamingCall{TRequest, TResponse}"/> representing the method call.</returns>
+        AsyncDuplexStreamingCall<TClientMessage, TServerMessage> Method(TClient client, CallOptions callOptions);
```
woksin: Maybe a name that's more familiar? Like Call?

Source/Services/IConvertReverseCallMessages.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// Defines a converter that reads and writes parts of duplex streaming messages using the reverse call protocol.
+    /// </summary>
+    /// <typeparam name="TClientMessage">Type of the messages that is sent from the client to the server.</typeparam>
+    /// <typeparam name="TServerMessage">Type of the messages that is sent from the server to the client.</typeparam>
+    /// <typeparam name="TConnectArguments">Type of the arguments that are sent along with the initial Connect call.</typeparam>
+    /// <typeparam name="TConnectResponse">Type of the response that is received after the initial Connect call.</typeparam>
+    /// <typeparam name="TRequest">Type of the requests sent from the server to the client.</typeparam>
+    /// <typeparam name="TResponse">Type of the responses received from the client.</typeparam>
+    public interface IConvertReverseCallMessages<TClientMessage, TServerMessage, TConnectArguments, TConnectResponse, TRequest, TResponse>
```
woksin, line 17: I think that we can think about this name. Doesn't really make sense considering what this represents

Source/Services/ICanCallADuplexStreamingMethod.cs
```diff
@@ -12,24 +12,17 @@ namespace Dolittle.SDK.Services
     /// <typeparam name="TClient">The type of generated gRPC client to use.</typeparam>
     /// <typeparam name="TClientMessage">Type of the <see cref="IMessage">messages</see> that is sent from the client to the server.</typeparam>
     /// <typeparam name="TServerMessage">Type of the <see cref="IMessage">messages</see> that is sent from the server to the client.</typeparam>
-    public interface IAmADuplexStreamingMethod<TClient, TClientMessage, TServerMessage>
+    public interface ICanCallADuplexStreamingMethod<TClient, TClientMessage, TServerMessage>
         where TClient : ClientBase<TClient>
```
woksin: You don't need the TClient anymore do you?

PR #6, opened 2020-09-10
Protobuf artifacts extensions

woksin:
Based off #3 

Adds extension methods for converting back and forth between Artifact and Protobuf Artifact.

I use some reflection magic in the .To<SomeArtifact>() method for correctly getting the instance of artifact that your looking for, with a couple of assumptions. But since the SDK most likely will own all the types derived from Artifact itself I don't think that this will become a problem. I think it is cool, and I certainly liked hacking it out :D


Source/Protobuf/ArtifactExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+                artifactIdValueProperty.SetValue(artifactId, artifact.Id.ToGuid());
+            }
+            catch (MissingMethodException)
+            {
+                throw new ArtifactIdDoesNotHaveDefaultParameterlessConstructor(typeof(TArtifact), artifact, idType);
+            }
+
+            Generation generation = artifact.Generation;
+            var constructor = typeof(TArtifact).GetConstructor(new[] { idType, generationType });
+            if (constructor == default) throw new ArtifactDoesNotHaveConstructorWithIdAndGeneration(typeof(TArtifact), artifact);
```
joelhoisko:
```suggestion
                throw new TypeDoesNotHaveDefaultParameterlessConstructor(typeof(TArtifact), artifact, idType);
            }

            Generation generation = artifact.Generation;
            var constructor = typeof(TArtifact).GetConstructor(new[] { idType, generationType });
            if (constructor == default) throw new TypeDoesNotHaveConstructorWithIdAndGeneration(typeof(TArtifact), artifact);
```

As this method is generalized I suggest we should drop the word `Artifact` from the exceptions, eg. calling `pbArtifact.To<EventType>()` and getting a `ArtifactDoesNotHaveConstructorWithIdAndGeneration` sounds a bit weird.
Not sure if the word `Type` is the correct replacement here, it's maybe a bit too vague.

Source/Protobuf/ArtifactDoesNotHaveConstructorWithIdAndGeneration.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class ArtifactDoesNotHaveConstructorWithIdAndGeneration : CouldNotConvertProtobufArtifact
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ArtifactDoesNotHaveConstructorWithIdAndGeneration"/> class.
+        /// </summary>
+        /// <param name="expectedArtifactType">The <see cref="Type" /> of <see cref="IArtifact" /> the protobuf artifact should be converted to.</param>
+        /// <param name="artifact">The <see cref="PbArtifact" />.</param>
+        public ArtifactDoesNotHaveConstructorWithIdAndGeneration(Type expectedArtifactType, PbArtifact artifact)
+            : base(expectedArtifactType, artifact, $"Artifact type {expectedArtifactType} does not have a constructor where the first parameter is a {typeof(ArtifactId)} and second parameter is a {typeof(Generation)}")
```
joelhoisko:
```suggestion
        public TypeDoesNotHaveConstructorWithIdAndGeneration(Type expectedType, PbArtifact artifact)
            : base(expectedArtifactType, artifact, $"Type {expectedType} does not have a constructor where the first parameter is a {typeof(ArtifactId)} and second parameter is a {typeof(Generation)}")
```

Same thing here where the given type isn't really an `Artifact`.

Source/Protobuf/ArtifactIdDoesNotHaveDefaultParameterlessConstructor.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using Dolittle.SDK.Artifacts;
+using PbArtifact = Dolittle.Artifacts.Contracts.Artifact;
+
+namespace Dolittle.SDK.Protobuf
+{
+    /// <summary>
+    /// Exception that gets thrown when an <see cref="ArtifactId" /> does not have a default parameterless constructor.
+    /// </summary>
+    public class ArtifactIdDoesNotHaveDefaultParameterlessConstructor : CouldNotConvertProtobufArtifact
```
joelhoisko: Same here with the `Artifact` naming but not quite sure what to call it here, `TypeId` sound very vague.

Source/Protobuf/CouldNotConvertProtobufArtifact.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class CouldNotConvertProtobufArtifact : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CouldNotConvertProtobufArtifact"/> class.
+        /// </summary>
+        /// <param name="expectedArtifactType">The <see cref="Type" /> of <see cref="IArtifact" /> the protobuf artifact should be converted to.</param>
+        /// <param name="artifact">The <see cref="PbArtifact" />.</param>
+        /// <param name="details">The details of the exception.</param>
+        public CouldNotConvertProtobufArtifact(Type expectedArtifactType, PbArtifact artifact, string details)
+            : base($"Could not convert artifact ({artifact.Id?.ToGuid()}, {artifact.Generation}) to artifact of type {expectedArtifactType}. {details}")
```
joelhoisko:
```suggestion
        public CouldNotConvertProtobufArtifact(Type expectedType, PbArtifact artifact, string details)
            : base($"Could not convert artifact ({artifact.Id?.ToGuid()}, {artifact.Generation}) to type {expectedType}. {details}")
```

Same here.

joelhoisko (changes requested):
Neato, some comments regarding the use of the word `Artifact`.

joelhoisko (approved):

PR #7, opened 2020-09-14
Event processing

woksin:
* Implements the different event processor kinds
* Specs for event processing request converter

This is not hooked up yet. We don't know if the things in Service work yet either, but I assume that the interface is not that far off. I'm also using a stub implementation of RetryPolicy as we need to implement something for resilience again. We'll swap that out when that's in place


Source/Events.Filters/Events.Filters.csproj
```diff
@@ -4,11 +4,14 @@
         <AssemblyName>Dolittle.SDK.Events.Filters</AssemblyName>
     </PropertyGroup>
     <ItemGroup>
+        <PackageReference Include="Dolittle.Runtime.Contracts" Version="$(ContractsVersion)" />
+        <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
```
joelhoisko:
```suggestion
        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
```

Source/Events.Handling/Events.Handling.csproj
```diff
@@ -4,11 +4,15 @@
         <AssemblyName>Dolittle.SDK.Events.Handling</AssemblyName>
     </PropertyGroup>
     <ItemGroup>
+        <PackageReference Include="Dolittle.Runtime.Contracts" Version="$(ContractsVersion)" />
+        <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
```
joelhoisko:
```suggestion
        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
```

Source/Events.Filters/Internal/EventFilterProcessor.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public FilterEventRequest GetRequestFrom(FilterRuntimeToClientMessage message)
+                => message.FilterRequest;
+
+            public void SetConnectArgumentsContextIn(ReverseCallArgumentsContext context, FilterRegistrationRequest arguments)
+                => arguments.CallContext = context;
+
+            public void SetResponseContextIn(ReverseCallResponseContext context, FilterResponse response)
+                => response.CallContext = context;
+        }
```
joelhoisko: We don't usually have nested classes so I'm not quite sure how to feel about this. The other option to make all of these `ICanCallADuplexStreamingMethod` and `IConvertReverseCallMessage` classes into their own files but I'm guessing we're scared of having too many small files around with their own docs.

For the converters would it be cleaner to make them into extensions?

Source/Events.Processing/MissingEventInformation.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Exception that gets thrown when there are details missing on an event.
+    /// </summary>
+    public class MissingEventInformation : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MissingEventInformation"/> class.
+        /// </summary>
+        /// <param name="details">The details on what is missing.</param>
+        public MissingEventInformation(string details)
+            : base($"Missing information on event: {details}")
```
joelhoisko:
```suggestion
            : base($"Event is missing information about its {details}")
```
Flows a bit nicer, when we have exceptions like these where the details are on a string I think it's nice that it flows nicely, kinda like how our tests also have the BDD-style descriptions to them.

Source/Events.Processing/EventTypeDoesNotMatchAssociatedEventType.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class EventTypeDoesNotMatchAssociatedEventType : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EventTypeDoesNotMatchAssociatedEventType"/> class.
+        /// </summary>
+        /// <param name="eventType">The <see cref="EventType" /> from the proceessing request.</param>
+        /// <param name="associatedEventType">The associated <see cref="EventType" />.</param>
+        /// <param name="type">The <see cref="Type" /> of the event.</param>
+        public EventTypeDoesNotMatchAssociatedEventType(EventType eventType, EventType associatedEventType, Type type)
+            : base($"Trying to convert event for prcessing to {type}, but {eventType} in the processing request does not match the associated artifact {associatedEventType}")
```
joelhoisko:
```suggestion
            : base($"Trying to convert event for processing to {type}, but {eventType} in the processing request does not match the associated artifact {associatedEventType}")
```

Source/Events.Processing/Events.Processing.csproj
```diff
@@ -3,11 +3,17 @@
     <PropertyGroup>
         <AssemblyName>Dolittle.SDK.Events.Processing</AssemblyName>
     </PropertyGroup>
+
     <ItemGroup>
+        <PackageReference Include="Dolittle.Runtime.Contracts" Version="$(ContractsVersion)" />
+        <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsVersion)" />
```
joelhoisko:
```suggestion
        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
```

Source/Events.Processing/IEventProcessingRequestConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using PbCommittedEvent = Dolittle.Runtime.Events.Contracts.CommittedEvent;
+using PbStreamEvent = Dolittle.Runtime.Events.Processing.Contracts.StreamEvent;
+
+namespace Dolittle.SDK.Events.Processing
+{
+    /// <summary>
+    /// Defines a system that can convert a request for processing an event to <see cref="EventContext" /> and the CLR instance of the event.
+    /// </summary>
+    public interface IEventProcessingRequestConverter
```
joelhoisko, line 13: Should it be `ICanConvertProcessingRequest`  like with the `IConvertReverseCallMessages` interfaces? Or is the `ICan` style reserved for interfaces that have multiple implementation?

Source/Events.Processing/IEventProcessingRequestConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /// <summary>
+        /// Gets the <see cref="EventContext" /> from the <see cref="PbCommittedEvent" />.
+        /// </summary>
+        /// <param name="event">The <see cref="PbCommittedEvent" />.</param>
+        /// <returns>The derived <see cref="EventContext" />.</returns>
+        EventContext GetEventContext(PbCommittedEvent @event);
```
joelhoisko, line 20:
```suggestion
        /// <summary>
        /// Converts a <see cref="PbCommittedEvent" /> to an <see cref="EventContext" />.
        /// </summary>
        /// <param name="event">The <see cref="PbCommittedEvent" />.</param>
        /// <returns>The derived <see cref="EventContext" />.</returns>
        EventContext ToEventContext(PbCommittedEvent @event);
```

As this is a converter shouldn't we be using the `ToSomething` wording here instead of `Get`? Kinda like how we had it with the extension methods? On the extension methods the `ToSomething` worked a bit nicer as it was extending the type but I think it still works here.

For maximum verbosity we could go for `ConvertProtobufEventToEventContext` or `ConvertToEventContext` or something along those lines (which I like neither off).

Source/Events/EventIdentifier.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (!BitConverter.IsLittleEndian)
+                Array.Reverse(sequenceNumberBytes);
+
+            using var stream = new MemoryStream(40);
+
+            stream.Write(microserviceBytes);
+            stream.Write(tenantBytes);
+            stream.Write(sequenceNumberBytes);
+
+            return Convert.ToBase64String(stream.ToArray());
```
joelhoisko, line 65: Are we sure that we don't accidentally mess with the casing of this string and break the base64 encoding? Asking just to be sure.

joelhoisko (changes requested):

Source/Events.Filters/Internal/EventFilterProcessor.cs
woksin: I think that these implementations are the exceptions to the rule. Since these implementations has no other purpose what so ever other than here. I would prefer not having these in their own files, but I don't know if I'm the only one thinking that? Thoughts, @jakhog @einari?

Source/Events.Processing/IEventProcessingRequestConverter.cs
woksin: We have usually used the ICan style for interfaces that naturally can have multiple implementations, as you said. This interface should only have one implementation, thus the name.

Source/Events.Processing/IEventProcessingRequestConverter.cs
woksin: Maybe "Converter" is the wrong name here. It does not really convert the message, it gets certain parts of it, for example the event context or the event itself. And given that context i don't think that `ToEventContext` makes sense

Source/Events/EventIdentifier.cs
woksin: This is just copy pasta code, @jakhog was the author. I don't even know if we need this anymore?

Source/Events.Processing/IEventProcessingRequestConverter.cs
joelhoisko: <s>So do we need the interface then anymore? With the old IoC stuff it made sense to have the interfaces but wouldn't the class be enough now? The whole class gets passed around anyways.</s>

Didn't have my morning coffee to think about this for longer than 2 secs hah

PR #8, opened 2020-09-15
Services specifications

jakhog:
Started with some tests for the Services (most notably the ReverseCallClient). Not complete yet in terms of testing all the features, and I found a nicer way of testing observables that we might want to use.

But I have fixed some interfaces and implementations that we might want to pull in now before other parts of the SDK go off track for too long.


woksin (approved):
Looks good

PR #10, opened 2020-09-17
Harmonize event converters

jakhog:
- Renamed EventProcessingRequestConverter -> EventProcessingConverter, and reused most of the logic from EventConverter
- Made non-throwing `Try...` variants of (almost) all Protobuf<->SDK conversion methods
- Re-implemented specifications for the happy paths of new converters


Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+                    source.EventLogSequenceNumber,
+                    source.Occurred.ToDateTimeOffset(),
+                    eventSourceId,
+                    executionContext,
+                    eventType,
+                    content,
+                    source.Public);
+                return true;
+            }
+        }
```
woksin: This is quite a long method. It could easily be split into several logical chunks.
For instance the first part validates the content of the source. Following the same pattern with Try... methods we could have a TryGetCommittedEventInformation(source, out var occurred, out var eventSourceId, ..., out error)

The deserialization part could be extracted out to a TryDeserializeEvent(source.Content, out content, out error)

The last if-branch could also be extracted out to a TryCreateCommittedEvent(..., out @event, out error) with all the arguments needed to create the SDK Committed Event :)

woksin: Also here should we check if (source == default) in the beginning perhaps?

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
-                    return new CommittedEvent(
-                        source.EventLogSequenceNumber,
-                        source.Occurred.ToDateTimeOffset(),
-                        source.EventSourceId.To<EventSourceId>(),
-                        source.ExecutionContext.ToExecutionContext(),
-                        eventType,
-                        content,
-                        source.Public);
+                    list.Add(@event);
                 }
```
woksin: I cannot suggest changes here for some reason but this part could be formatted in another way:

```
if (!TryToSDK(sourceEvent), out var @event, out error)) return false;
list.Add(@event);
```
Or am I missing something?

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+
+        /// <inheritdoc/>
+        public bool TryToSDK(Contracts.CommitEventsResponse source, out CommitEventsResult result, out Exception error)
+        {
+            result = default;
+
+            if (!TryToSDK(source.Events, out var events, out error))
+            {
+                return false;
+            }
```
woksin: I don't know, but I just like having return something; on the same line of the if-statement if that's the only thing it does
```suggestion
            if (!TryToSDK(source.Events, out var events, out error)) return false;
```

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+
+            error = null;
```
woksin: This line is unnecessary since if TryToSDK is false, then the error should be null given the expected behaviour of the Try-methods
```suggestion
```

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+                },
+                Formatting = Formatting.None,
+            };
+            content = JsonConvert.SerializeObject(source.Content, serializerSettings);
+
+            if (serializationFailed)
+            {
+                error = new CouldNotSerializeEventContent(source.Content, serializationError);
+                return false;
+            }
```
woksin: This could be a extracted out to a TrySerializeEvent-method

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+
+            error = null;
```
woksin, line 177: If we do as I suggested above we could invoke TrySerializeEvent(source.Content, out var content, out error) and then we wouldn't need to have this statement
```suggestion
```

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+            foreach (var sourceEvent in source)
+            {
+                if (!TryToProtobuf(sourceEvent, out var @event, out error))
+                {
+                    return false;
+                }
+                else
+                {
+                    list.Add(@event);
+                }
```
woksin: Same comment as above

Source/Events/EventConverter.cs
```diff
@@ -27,69 +26,231 @@ public EventConverter(IEventTypes eventTypes)
...
+            }
+
+            error = null;
```
woksin, line 212: Same comment as above
```suggestion
```

Source/Protobuf/ArtifactExtensions.cs
```diff
@@ -28,41 +28,99 @@ public static PbArtifact ToProtobuf(this IArtifact artifact)
...
+        /// <param name="artifact">When the method returns, a tuple with <see cref="ArtifactId" /> and <see cref="Generation" /> if conversion was successful, otherwise null.</param>
+        /// <param name="error">When the method returns, null if the conversion was successful, otherwise the error that caused the failure.</param>
+        /// <returns>A value indicating whether or not the conversion was successful.</returns>
+        public static bool TryToArtifact(this PbArtifact source, out (ArtifactId Id, Generation Generation) artifact, out Exception error)
+        {
+            artifact = default;
+            if (!source.Id.TryTo<ArtifactId>(out var id, out error))
+            {
+                return false;
+            }
```
woksin: As I've mentioned above, I like the return something on the same line. Up to you to decide :) 
```suggestion
            if (!source.Id.TryTo<ArtifactId>(out var id, out error)) return false;
```

Source/Protobuf/ArtifactExtensions.cs
```diff
@@ -28,41 +28,99 @@ public static PbArtifact ToProtobuf(this IArtifact artifact)
...
 
-            var instance = constructor.Invoke(new object[] { artifactId, generation });
-            if (instance == default) throw new CouldNotConvertProtobufArtifact(typeof(TArtifact), artifact, "Could not create instance of artifact");
-            return instance as TArtifact;
+            artifact = (id, source.Generation);
```
woksin, line 112: What if source.Generation is null?

Source/Protobuf/ArtifactExtensions.cs
```diff
@@ -28,41 +28,99 @@ public static PbArtifact ToProtobuf(this IArtifact artifact)
...
+                }
+
+                artifact = constructor.Invoke(new object[] { artifactId, (Generation)source.Generation }) as TArtifact;
+                if (artifact == default)
+                {
+                    error = new CouldNotConvertProtobufArtifact(typeof(TArtifact), source, "Could not create instance of artifact");
+                    return false;
+                }
+
+                error = null;
```
woksin, line 69: Could make this method smaller by extracting this part out to a TryCreateArtifact or something along those lines.

Source/Protobuf/ClaimExtensions.cs
```diff
@@ -39,8 +40,48 @@ public static Claim ToClaim(this PbClaim claim)
...
+                    return false;
+                }
+
+                if (string.IsNullOrEmpty(claim.Key))
+                {
+                    error = new InvalidClaimsConversion("one of the claims was missing the key");
+                    return false;
+                }
+
+                list.Add(new Claim(claim.Key, claim.Value, claim.ValueType));
```
woksin, line 71: Why don't we have a TryToClaim when we have TryToClaims? Could use that TryToClaim here for a shorter and sweeter method :)

Source/Protobuf/ExecutionContextExtensions.cs
```diff
@@ -36,16 +37,62 @@ public static PbExecutionContext ToProtobuf(this ExecutionContext executionConte
...
+                microserviceId,
+                tenantId,
+                source.Version.ToVersion(),
+                source.Environment,
+                correlationId,
+                claims,
                 CultureInfo.InvariantCulture);
+            error = null;
+            return true;
+        }
```
woksin, line 88: In the spirit of my previous comments.. Long methods. Could split this into chunks as well? As a general thing I do prefer having more smaller methods than fewer longer, but of course within the realms of sanity. It shoud make sense in the given context :P

Source/Protobuf/Failure.cs
```diff
@@ -36,18 +36,37 @@ public Failure(FailureId id, FailureReason reason)
...
-        public static implicit operator Contracts.Failure(Failure failure) =>
-            failure != null ?
-                new Contracts.Failure { Id = failure.Id.ToProtobuf(), Reason = failure.Reason }
-                : null;
+        public static implicit operator Contracts.Failure(Failure failure)
+        {
+            if (failure == null)
+            {
+                return null;
+            }
```
woksin: See previous comments on similar suggestion
```suggestion
            if (failure == null) return null;
```

Source/Protobuf/Failure.cs
```diff
@@ -36,18 +36,37 @@ public Failure(FailureId id, FailureReason reason)
...
-        public static implicit operator Failure(Contracts.Failure failure) =>
-            failure != null ?
-                new Failure(failure.Id.To<FailureId>(), failure.Reason)
-                : null;
+        public static implicit operator Failure(Contracts.Failure failure)
+        {
+            if (failure == null)
+            {
+                return null;
+            }
```
woksin: See above
```suggestion
            if (failure == null) return null;
```

woksin (changes requested):
Good stuff, I like the Try-method stuff for avoiding throwing exceptions all over the place.

I do have a few comments which are related to formating and code-style. I don't have really strong opinions on this, but I do think that we should try to be consistent in style across the board.

Source/Events/EventConverter.cs
woksin: I see that I forgot to include the top part of the method

Source/Events/EventConverter.cs
woksin: I meant to say if TryToSDK is true

Source/Protobuf/ArtifactExtensions.cs
jakhog: It cannot be null :)

woksin: Ohh, yes it cannot :P

Source/Protobuf/ArtifactExtensions.cs
jakhog: This one could probably be simplified yes - but honestly I was a bit dumbfounded when I met this code, and just tried to make it work. The whole generic artifact thing is somewhat obscure to me still.

Source/Events/EventConverter.cs
jakhog: I don't mind either way, but I honestly thought we had C# coding style guidelines that said we should have curly braces for all the ifs. But it seems that was just in my head.

woksin: I do remember talking about it. I just have a personal preference for those one liners with only a return

Could perhaps get some input from The C# man @einari

Source/Protobuf/ArtifactExtensions.cs
woksin: I wrote this, so I think it makes perfect sense given that we want to do it this way :P We could drop having this and instead have a ToSomeType for every artifact type we have. For instance ToEventType

Source/Events/EventConverter.cs
einari: I tend to use oneliners, if it is readable - typically very short ones.

Don't see any harm in doing that. Readability is the key here.

Source/Events/EventConverter.cs
einari: Just got an idea here, might not make sense. 

But what if you had an inner class called ToSDK and one that was called ToProtobuf, and on it you had Try() methods. 

Then it would read like this:

```csharp
ToSDK.Try(....)
```

einari:
Consistency is important.

Quoting John Wood from 1991: 

[...] Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live. Code for readability.

https://books.google.co.id/books?id=gJf9tI2mytIC&pg=PT27#v=snippet&q=%22Always%20code%20as%20if%20the%20guy%20who%20ends%20up%20maintaining%20your%20code%20will%20be%20a%20violent%20psychopath%20who%20knows%20where%20you%20live%22&f=false

Source/Events/EventConverter.cs
woksin: I don't really see the benefits with that approach. Care to elaborate?

Source/Events/EventConverter.cs
jakhog: Moved the serializer parts out to it's own methods - and checked for nulls. Moving the last if-branch didn't really help making it any nicer to look at, so I skipped that. Perphaps something else we will change about that two-typed thing later.

Source/Events/EventConverter.cs
jakhog: Done

Source/Events/EventConverter.cs
jakhog: Done

Source/Events/EventConverter.cs
jakhog: Done

Source/Events/EventConverter.cs
jakhog: Done

Source/Events/EventConverter.cs
jakhog: Done

Source/Protobuf/ArtifactExtensions.cs
jakhog: Done

Source/Protobuf/ArtifactExtensions.cs
jakhog: I'll leave this for later - as I think the Artifact things might need another lookover. I'm a bit freaked out when we need to to reflection-based things to make something this fundamental work.

Source/Protobuf/ClaimExtensions.cs
jakhog: Actually - I would rather like to remove the methods for single Claim objects, as we never use them anywhere else in the code...

Source/Protobuf/Failure.cs
jakhog: Done

Source/Protobuf/Failure.cs
jakhog: Done

Source/Protobuf/ExecutionContextExtensions.cs
jakhog: I don't really see how to make this much shorter?

Source/Events/EventConverter.cs
jakhog: Done

jakhog:
I've addressed most of your comments - but some still left for discussion.

PR #11, opened 2020-09-17
Events handling

woksin:
Makes it possible to build event handlers


Source/Events.Handling/Builder/IEventHandlerMethod.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Invokes the event handler method.
+        /// </summary>
+        /// <param name="event">The event.</param>
+        /// <param name="context">The <see cref="EventContext" />.</param>
+        /// <returns>A <see cref="Task" /> representing the asynchronous action.</returns>
+        Task Invoke(object @event, EventContext context);
+    }
+}
```
jakhog: I am wondering why we need both the delegates and this interface (whey they do the same thing). Does it have something to do with what comes next for the EventHandler classes?

Source/Events.Handling/Builder/EventHandlerBuilder.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        /// <summary>
+        /// Build event handler methods.
+        /// </summary>
+        /// <param name="callback">The callback for building event handler methods.</param>
+        public void WithMethods(Action<EventHandlerMethodsBuilder> callback)
+        {
+            _methodsBuilder = new EventHandlerMethodsBuilder(_eventHandlerId, _loggerFactory.CreateLogger<EventHandlerMethodsBuilder>());
+            callback(_methodsBuilder);
+        }
```
jakhog, line 77: Personally - I think this extra layer of builder callback is not very pleasing on the eye. And I don't see any reason why we _need_ it?

Source/Events.Handling/Builder/TypedEventHandlerMethod.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <param name="context">The <see cref="EventContext" />.</param>
+        /// <returns>A <see cref="Task" /> representing the asynchronous action.</returns>
+        public Task Invoke(T @event, EventContext context) => _method(@event, context);
+
+        /// <inheritdoc/>
+        Task IEventHandlerMethod.Invoke(object @event, EventContext context)
+        {
+            if (@event is T) return Invoke(@event as T, context);
+            throw new TypedEventHandlerMethodInvokedOnEventOfWrongType(_eventHandlerId, typeof(T), @event.GetType());
+        }
```
jakhog: Wouldn't it look nicer to just invoke the method directly in the implemented interface method?
```suggestion
        /// <inheritdoc/>
        Task Invoke(object @event, EventContext context)
        {
            if (@event is T typedEvent) return _method(typedEvent, context);
            throw new TypedEventHandlerMethodInvokedOnEventOfWrongType(_eventHandlerId, typeof(T), @event.GetType());
        }
```

jakhog (changes requested):

Source/Events.Handling/Builder/EventHandlerBuilder.cs
woksin: This was done to be able to create multiple event handler methods for an event handler. I'm open to suggestions on how to do that in a less layered way

Source/Events.Handling/Builder/TypedEventHandlerMethod.cs
woksin: Yes it would 🤣

Source/Events.Handling/Builder/IEventHandlerMethod.cs
woksin: The delegates represents the two different ways of handling an event (basically either a specific event or an object event) and the IEventHandlerMethod defines an interface that works for both of them. Without having it I don't know how we would achieve the EventHandler class technically because one of the two event handler delegates is generic. If you have suggestions I would like to hear about them :)

Source/Events.Handling/Builder/EventHandlerBuilder.cs
jakhog: We will look at this compared to the guide and TS SDK later.

jakhog (approved):

PR #13, opened 2020-09-22
Reverse call specifications

jakhog:
- Tried to clarify the internal logic of the ReverseCallClient by splitting the main body into methods.
- ReverseCallClient now requires an `IScheduler` for the reactive processing, this is needed mostly for testing, and for normal execution we use the TaskPoolScheduler (provided by the ReverseCallClientCreator).
- Used Microsoft.Reactive.Testing framework for testing of the ReverseCallClient, this simplifies verifying the behaviour of observables without actually running code and waiting for it.
- New tests for the request handling part of the ReverseCallClient


woksin (changes requested):
Looks good to me 👍

woksin (approved):
Whoops, meant to approve :)

PR #14, opened 2020-09-22
Reintroduce specs for Concepts + create specs for event handler processor

woksin:
Also fix a couple of typos and reformat some code in Concepts with lambdas => and switch expressions with pattern matching


Source/Concepts/ConceptAs.cs
```diff
@@ -51,54 +51,39 @@ public class ConceptAs<T> : IEquatable<ConceptAs<T>>, IComparable<ConceptAs<T>>,
...
         /// Larger than comparison operator for comparing two <see cref="ConceptAs{T}"/>.
         /// </summary>
         /// <param name="left">Left <see cref="ConceptAs{T}"/>.</param>
         /// <param name="right">Right <see cref="ConceptAs{T}"/>.</param>
         /// <returns>true if the left <see cref="ConceptAs{T}"/> is larger than the right <see cref="ConceptAs{T}"/>; otherwise, false.</returns>
-        public static bool operator >(ConceptAs<T> left, ConceptAs<T> right)
-        {
-            return left.CompareTo(right) == 1;
-        }
+        public static bool operator >(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) == 1;
```
jakhog: See you haven't changed this - but I'm pretty sure it should be:
```suggestion
        public static bool operator >(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) > 0;
```

Source/Concepts/ConceptAs.cs
```diff
@@ -51,54 +51,39 @@ public class ConceptAs<T> : IEquatable<ConceptAs<T>>, IComparable<ConceptAs<T>>,
...
         /// Smaller than comparison operator for comparing two <see cref="ConceptAs{T}"/>.
         /// </summary>
         /// <param name="left">Left <see cref="ConceptAs{T}"/>.</param>
         /// <param name="right">Right <see cref="ConceptAs{T}"/>.</param>
         /// <returns>true if the left <see cref="ConceptAs{T}"/> is smaller than the right <see cref="ConceptAs{T}"/>; otherwise, false.</returns>
-        public static bool operator <(ConceptAs<T> left, ConceptAs<T> right)
-        {
-            return left.CompareTo(right) == -1;
-        }
+        public static bool operator <(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) == -1;
```
jakhog: Same as above
```suggestion
        public static bool operator <(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) < 0;
```

Source/Concepts/ConceptAs.cs
```diff
@@ -51,54 +51,39 @@ public class ConceptAs<T> : IEquatable<ConceptAs<T>>, IComparable<ConceptAs<T>>,
...
         /// Larger than or equal comparison operator for comparing two <see cref="ConceptAs{T}"/>.
         /// </summary>
         /// <param name="left">Left <see cref="ConceptAs{T}"/>.</param>
         /// <param name="right">Right <see cref="ConceptAs{T}"/>.</param>
         /// <returns>true if the left <see cref="ConceptAs{T}"/> is larger or equal than the right <see cref="ConceptAs{T}"/>; otherwise, false.</returns>
-        public static bool operator >=(ConceptAs<T> left, ConceptAs<T> right)
-        {
-            return left.CompareTo(right) > -1;
-        }
+        public static bool operator >=(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) > -1;
```
jakhog: This doesn't matter - but I just like the use of 0 better.
```suggestion
        public static bool operator >=(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) >= 0;
```

Source/Concepts/ConceptAs.cs
```diff
@@ -51,54 +51,39 @@ public class ConceptAs<T> : IEquatable<ConceptAs<T>>, IComparable<ConceptAs<T>>,
...
         /// Smaller or equal than comparison operator for comparing two <see cref="ConceptAs{T}"/>.
         /// </summary>
         /// <param name="left">Left <see cref="ConceptAs{T}"/>.</param>
         /// <param name="right">Right <see cref="ConceptAs{T}"/>.</param>
         /// <returns>true if the left <see cref="ConceptAs{T}"/> is smaller or equal than the right <see cref="ConceptAs{T}"/>; otherwise, false.</returns>
-        public static bool operator <=(ConceptAs<T> left, ConceptAs<T> right)
-        {
-            return left.CompareTo(right) < 1;
-        }
+        public static bool operator <=(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) < 1;
```
jakhog: Same as above
```suggestion
        public static bool operator <=(ConceptAs<T> left, ConceptAs<T> right) => left.CompareTo(right) <= 0;
```

Specifications/Concepts/for_ConceptAs/given/comparable_concepts.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Dolittle.SDK.Concepts.for_ConceptAs.given
+{
+    public class comparable_concepts : test_concepts
+    {
+        public const int LESS_THAN = -1;
+        public const int EQUAL = 0;
+        public const int GREATER_THAN = 1;
```
jakhog, line 10: Regarding my comments above - these might not be correct for all types. My thinking is based on this https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.comparer-1.compare?view=netcore-3.1 

But it might not be important for the types of concepts we are testing here.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/given/some_event.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Dolittle.SDK.Concepts;
+
+namespace Dolittle.SDK.Events.Handling.Internal.for_EventHandlerProcessor.given
+{
+    public class some_event : Value<some_event>
+    {
+        public string some_string { get; set; }
+    }
```
jakhog, line 11: We don't require that events are value types, is there a particular reason why it is here?

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_handler_does_not_fail.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class and_handler_does_not_fail : given.all_dependencies
+    {
+        static EventHandlerResponse response;
+
+        Establish context = () => event_handler.Setup(_ => _.Handle(Moq.It.IsAny<object>(), Moq.It.IsAny<EventType>(), Moq.It.IsAny<EventContext>(), Moq.It.IsAny<CancellationToken>())).Returns(Task.CompletedTask);
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_not_have_a_failure = () => response.Failure.ShouldBeNull();
```
jakhog: Does it make sense to also check that the EventHandlerProcessor calls `Handle` with the correct arguments, namely the correct `request` and `execution_context`.

You can use `event_handler.Verify(...)` for that.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_handler_fails.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        static EventHandlerResponse response;
+
+        Establish context = () => event_handler.Setup(_ => _.Handle(Moq.It.IsAny<object>(), Moq.It.IsAny<EventType>(), Moq.It.IsAny<EventContext>(), Moq.It.IsAny<CancellationToken>())).Throws(new System.Exception());
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
```
jakhog, line 52: Same here - about checking the arguments.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_empty_event_content.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.Event.Content = string.Empty;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: This test seem to be the responsibility of the `EventConverter` to deal with - so maybe it should rather test what happens if the `EventConverter.ToSDK(...)` throws an exception?

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_event_that_cannot_be_deserialized.cs
```diff
@@ -0,0 +1,29 @@
...
+            request.Event.Event.Content = "{\"something\": \"hello\"";
+        };
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: If you change the previous test to check for exceptions from the `EventConverter`, it will cover both of these cases.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_missing_event_source_id.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.Event.EventSourceId = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_missing_event_type.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.Event.Type = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_missing_execution_context.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.Event.ExecutionContext = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_missing_occurred_time_stamp.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.Event.Occurred = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/is_missing_committed_event.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.Event = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/is_missing_partition_id.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.PartitionId = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/is_missing_scope_id.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        Establish context = () => request.Event.ScopeId = null;
+
+        Because of = () => response = event_handler_processor.Handle(request, execution_context, CancellationToken.None).GetAwaiter().GetResult();
+
+        It should_get_a_response = () => response.ShouldNotBeNull();
+        It should_have_a_failure = () => response.Failure.ShouldNotBeNull();
+        It should_try_to_retry_processing = () => response.Failure.Retry.ShouldBeTrue();
+    }
+}
```
jakhog: Same here. If these are not covered by the EventConverter specs, it should be moved there.

Specifications/Events.Handling/for_EventHandler/when_handling/and_handler_method_fails.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        static Exception exception;
+
+        Establish context = ()
+            => event_handler_method.Setup(_ => _.TryHandle(Moq.It.IsAny<object>(), Moq.It.IsAny<EventContext>())).Returns(Task.FromResult<Try>(new Exception("Something went wrong")));
+
+        Because of = () => exception = Catch.Exception(() => event_handler.Handle(@event, handled_event_type, event_context, CancellationToken.None).GetAwaiter().GetResult());
+
+        It should_invoke_handler_with_correct_arguments = () => event_handler_method.Verify(_ => _.TryHandle(@event, event_context), Moq.Times.Once);
+        It should_throw_exception_because_event_handler_method_failed = () => exception.ShouldBeOfExactType<EventHandlerMethodFailed>();
```
jakhog: We could also check that it carries along the correct inner exception:
```suggestion
        static Exception thrownException;
        static Exception exception;

        Establish context = () => 
        {
            thrownException = new Exception("Something went wrong");
            event_handler_method
                .Setup(_ => _.TryHandle(Moq.It.IsAny<object>(), Moq.It.IsAny<EventContext>()))
                .Returns(Task.FromResult<Try>(thrownException));
        }

        Because of = () => exception = Catch.Exception(() => event_handler.Handle(@event, handled_event_type, event_context, CancellationToken.None).GetAwaiter().GetResult());

        It should_invoke_handler_with_correct_arguments = () => event_handler_method.Verify(_ => _.TryHandle(@event, event_context), Moq.Times.Once);
        It should_throw_exception_because_event_handler_method_failed = () => exception.ShouldBeOfExactType<EventHandlerMethodFailed>();
        It should_throw_exception_that_contains_the_original_thrown_exception = () => exception.InnerException.ShouldEqual(thrownException);
```
Sorry if this doesn't build...

Specifications/Events.Handling/for_EventHandler/when_handling/an_event.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class an_event : given.all_dependencies
+    {
+        static Exception exception;
+
+        Establish context = ()
+            => event_handler_method.Setup(_ => _.TryHandle(Moq.It.IsAny<object>(), Moq.It.IsAny<EventContext>())).Returns(Task.FromResult(new Try()));
+
+        Because of = () => exception = Catch.Exception(() => event_handler.Handle(@event, handled_event_type, event_context, CancellationToken.None).GetAwaiter().GetResult());
+
+        It should_invoke_handler_with_correct_arguments = () => event_handler_method.Verify(_ => _.TryHandle(@event, event_context), Moq.Times.Once);
```
jakhog, line 21: Since we easily can - should we check that it doesn't call the event handler method more? With the `.VerifyNoOtherCalls()`.

jakhog (changes requested):
Some comments directly in code, but as a bigger more general thing; the specifications around comparisons of `ConceptAs<...>` are quite big. The best practice is to only have single-line `Because` statements. In these tests we could also just do the checks in the `It` statements directly, so we don't need to store a million bools.

Also, I'm wondering if there is a way to simplify them. In the end, all that we care about is that for a comparison the `ConceptAs<...>` calls the underlying comparator. Then we don't need to care that it works. But I think it is easier to discuss and try out some things live together for these questions.

Specifications/Concepts/for_ConceptAs/given/comparable_concepts.cs
woksin: I agree with you. But as you may have noticed, these specs are just directly copied over from DotNET.Fundamentals.

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/given/some_event.cs
woksin: Yeah I can drop that here, I did it like this for another spec to be able to have an `It`test where I can check the equality between two events without having to check each property

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_handler_does_not_fail.cs
woksin: Yes that makes sense, I should do that

Specifications/Events.Handling/Internal/for_EventHandlerProcessor/when_handling/a_request_to_handle_event/and_stream_event/has_committed_event/with_empty_event_content.cs
woksin: True, I didn't think about mocking out the concrete EventProcessingConverter class. I'll just do that then.

jakhog (commented):
I just have some minor things - and the suggestion wouldn't work here. So I'll just make a commit and then you can have a look and see if you agree 🙂

jakhog (approved):

jakhog:
@woksin just have a look at the last changes I did, and if you agree merge it in

PR #15, opened 2020-09-22
Eventstore specs

joelhoisko:
- Test for all the different public methods on `EventStore`
- I decided to mock away the converter (we already have tests for it), comments if it's too crude.
- Also added a spec for `CommitEventsResult EventConverter.ToSDK(Contracts.CommitEventsResponse)`.


Specifications/Events/for_EventConverter/when_converting_a_committed_event_to_sdk/from_a_commit_events_response.cs
```diff
@@ -0,0 +1,96 @@
...
+using Dolittle.SDK.Protobuf;
+using Google.Protobuf.WellKnownTypes;
+using Machine.Specifications;
+using Newtonsoft.Json;
+using PbArtifact = Dolittle.Artifacts.Contracts.Artifact;
+using PbCommittedEvent = Dolittle.Runtime.Events.Contracts.CommittedEvent;
+
+namespace Dolittle.SDK.Events.for_EventConverter.when_converting_a_committed_event_to_sdk
+{
+    public class from_a_commit_events_response : given.a_converter_and_a_protobuf_execution_context
```
jakhog: Rename the test to:
```suggestion
namespace Dolittle.SDK.Events.for_EventConverter
{
    public class when_converting_a_commit_events_response_to_sdk : given.a_converter_and_a_protobuf_execution_context
```
Also move the file to match the name

Specifications/Events/for_EventStore/given/an_event_store_and_an_execution_context.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            // we don't care here what the converter is doing, just return the correct type at least
+            converter.Setup(_ => _.ToSDK(Moq.It.IsAny<Contracts.CommitEventsResponse>()))
+                .Returns(Moq.It.IsAny<CommitEventsResult>);
+
+            // allows us to inspect the uncommitted events just before they'd get committed and make it
+            // return the correct type and amount at least
+            converter.Setup(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()))
+                .Callback<UncommittedEvents>((events) => events_to_convert = events)
+                .Returns((UncommittedEvents @events) => @events.Select(_ => new Contracts.UncommittedEvent { }));
```
jakhog: A few things here:

1. There shouldn't be a need for comments in the code. If it is unclear what the code does - it's better to wrap it into a method with a more descriptive name.
2. The `.Returns(Moq...)` on line 34 doesn't work the way you seem to think. It will return the specific thing that you provide - and frankly I'm a bit surprised that this compiles 😅
3. I think you are making your life a bit harder than it needs to be here with returning actual results and tracking the things it sends in to the converter.

I think the tests would be simplified if you just return a specific object (so `CommitEventsResult` and `UncommittedEvents`) from the converter, and check that the `EventStore` implementation passes them along untouched to the other things it uses.

Specifications/Events/for_EventStore/when_committing_a_public_object.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Dolittle.SDK.Events.for_EventConverter.given;
+using Machine.Specifications;
+using It = Machine.Specifications.It;
+
+namespace Dolittle.SDK.Events.for_EventStore
+{
+    public class when_committing_a_public_object : given.an_event_store_and_an_execution_context
```
jakhog:
```suggestion
    public class when_committing_a_public_event_with_known_type : given.an_event_store_and_an_execution_context
```

Specifications/Events/for_EventStore/when_committing_a_public_object.cs
```diff
@@ -0,0 +1,33 @@
...
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye public world", 12345, true);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
+
+            event_types.Setup(_ => _.GetFor(content.GetType())).Returns(event_type);
```
jakhog: It seems like there should also be a test for when the `event_types` doesn't know about the type?

Specifications/Events/for_EventStore/when_committing_a_public_object_with_an_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Dolittle.SDK.Events.for_EventConverter.given;
+using Machine.Specifications;
+using It = Machine.Specifications.It;
+
+namespace Dolittle.SDK.Events.for_EventStore
+{
+    public class when_committing_a_public_object_with_an_event_type : given.an_event_store_and_an_execution_context
```
jakhog:
```suggestion
    public class when_committing_a_public_event_with_an_event_type : given.an_event_store_and_an_execution_context
```

Specifications/Events/for_EventStore/when_committing_a_public_object_with_an_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+        static an_event content;
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye public world", 12345, true);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
+        };
```
jakhog: To be thorough, we could also check that if you send in an event of type that has been associated with an `EventType`, it doesn't use the registered one, but rather the one you provide as an argument.

Specifications/Events/for_EventStore/when_committing_a_public_object.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        static an_event content;
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye public world", 12345, true);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
```
jakhog: Use hardcoded random guids as values.

Specifications/Events/for_EventStore/when_committing_a_public_object_with_an_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        static an_event content;
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye public world", 12345, true);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
```
jakhog: Use hardcoded random guids as values.

Specifications/Events/for_EventStore/when_committing_an_object.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Dolittle.SDK.Events.for_EventConverter.given;
+using Machine.Specifications;
+using It = Machine.Specifications.It;
+
+namespace Dolittle.SDK.Events.for_EventStore
+{
+    public class when_committing_an_object : given.an_event_store_and_an_execution_context
```
jakhog:
```suggestion
    public class when_committing_an_event: given.an_event_store_and_an_execution_context
```

Specifications/Events/for_EventStore/when_committing_an_object.cs
```diff
@@ -0,0 +1,33 @@
...
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye world", 12345, true);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
+
+            event_types.Setup(_ => _.GetFor(content.GetType())).Returns(event_type);
```
jakhog: Same here - we should probably specify what should happen if it does not know about the type.

Specifications/Events/for_EventStore/when_committing_an_object_with_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Dolittle.SDK.Events.for_EventConverter.given;
+using Machine.Specifications;
+using It = Machine.Specifications.It;
+
+namespace Dolittle.SDK.Events.for_EventStore
+{
+    public class when_committing_an_object_with_event_type : given.an_event_store_and_an_execution_context
```
jakhog:
```suggestion
    public class when_committing_an_event_with_event_type : given.an_event_store_and_an_execution_context
```

Specifications/Events/for_EventStore/when_committing_an_object.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        static an_event content;
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye world", 12345, true);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
```
jakhog: Use hardcoded random guids as values.

Specifications/Events/for_EventStore/when_committing_an_object_with_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        static an_event content;
+        static EventSourceId event_source;
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye world", 12345, false);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
```
jakhog: Use hardcoded random guids as values.

Specifications/Events/for_EventStore/when_committing_an_uncommitted_event.cs
```diff
@@ -0,0 +1,35 @@
...
+        static an_event content;
+        static EventSourceId event_source;
+        static EventType event_type;
+        static bool is_public;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye world", 12345, false);
+            event_source = new EventSourceId(Guid.NewGuid());
+            event_type = new EventType(Guid.NewGuid());
```
jakhog: Use hardcoded random guids as values.

Specifications/Events/for_EventStore/when_committing_an_uncommitted_event.cs
```diff
@@ -0,0 +1,35 @@
...
+            is_public = false;
+            uncommitted_event = new UncommittedEvent(event_source, event_type, content, is_public);
+        };
+
+        Because of = async () => await event_store.Commit(uncommitted_event);
+        It should_convert_the_object_to_uncommitted_events = () => events_to_convert.ShouldBeOfExactType<UncommittedEvents>();
```
jakhog: It should be enough to check that it tries to convert the specific `uncommitted_event`. So something like:
```csharp
It should_convert_the_provided_event = () => events_to_convert.ShouldContainOnly(uncommitted_event);
```

Specifications/Events/for_EventStore/when_committing_an_uncommitted_event.cs
```diff
@@ -0,0 +1,35 @@
...
+        It should_have_the_correct_event_source_id = () => events_to_convert[0].EventSource.ShouldEqual(uncommitted_event.EventSource);
+        It should_have_the_correct_event_type = () => events_to_convert[0].EventType.ShouldEqual(uncommitted_event.EventType);
+        It should_have_the_correct_content = () => events_to_convert[0].Content.ShouldEqual(uncommitted_event.Content);
+        It shouldnt_be_public = () => events_to_convert[0].IsPublic.ShouldEqual(uncommitted_event.IsPublic);
+    }
```
jakhog, line 39: The most important check is also to check that it sends the `CommitEventsRequest` with the correct call context and the results from the converter to the caller.

Specifications/Events/for_EventStore/when_committing_uncommitted_events.cs
```diff
@@ -0,0 +1,62 @@
...
+            uncommitted_events = new UncommittedEvents
+            {
+                event_one,
+                event_two,
+                event_two,
+            };
+        };
+
+        Because of = async () => await event_store.Commit(uncommitted_events);
+        It should_have_called_the_caller = () => caller.Verify(_ => _.Call(Moq.It.IsAny<EventStoreCommitMethod>(), Moq.It.IsAny<Contracts.CommitEventsRequest>(), Moq.It.IsAny<CancellationToken>()));
```
jakhog: Verify that the caller is called with the correct arguments - and also that it is not called for anything else (so we don't commit twice).

Specifications/Events/for_EventStore/when_committing_uncommitted_events.cs
```diff
@@ -0,0 +1,62 @@
...
+        It should_have_added_the_execution_context_to_the_call_context = () => commit_events_request.CallContext.ExecutionContext.ShouldEqual(execution_context.ToProtobuf());
+        It should_have_added_a_not_set_head_id_to_the_call_context = () => commit_events_request.CallContext.HeadId.ShouldEqual(HeadId.NotSet.Value.ToProtobuf());
```
jakhog: Since the `HeadId` is not set to a useful value right now - I don't think there is any need to specify that.

Specifications/Events/for_EventStore/when_committing_uncommitted_events.cs
```diff
@@ -0,0 +1,62 @@
...
+        It should_have_the_same_content_for_the_second_event = () => events_to_convert[1].Content.ShouldEqual(uncommitted_events[1].Content);
+        It should_have_the_same_is_public_for_the_second_event = () => events_to_convert[1].IsPublic.ShouldEqual(uncommitted_events[1].IsPublic);
+        It should_have_the_same_event_source_for_the_third_event = () => events_to_convert[2].EventSource.ShouldEqual(uncommitted_events[2].EventSource);
+        It should_have_the_same_event_type_for_the_third_event = () => events_to_convert[2].EventType.ShouldEqual(uncommitted_events[2].EventType);
+        It should_have_the_same_event_type_generation_for_the_third_event = () => events_to_convert[2].EventType.Generation.ShouldEqual(uncommitted_events[2].EventType.Generation);
+        It should_have_the_same_content_for_the_third_event = () => events_to_convert[2].Content.ShouldEqual(uncommitted_events[2].Content);
+        It should_have_the_same_is_public_for_the_third_event = () => events_to_convert[2].IsPublic.ShouldEqual(uncommitted_events[2].IsPublic);
```
jakhog: Same here, just check that it calls the converter with the right events, and that it uses the results from the converter to call the caller.

Specifications/Events/for_EventStore/when_committing_an_object_with_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+        };
+
+        Because of = async () => await event_store.Commit(content, event_source, event_type);
+        It should_convert_the_object_to_uncommitted_events = () => events_to_convert.ShouldBeOfExactType<UncommittedEvents>();
+        It should_have_the_correct_event_source_id = () => events_to_convert[0].EventSource.ShouldEqual(event_source);
+        It should_have_the_correct_event_type = () => events_to_convert[0].EventType.ShouldEqual(event_type);
+        It should_have_the_correct_content = () => events_to_convert[0].Content.ShouldEqual(content);
+        It shouldnt_be_public = () => events_to_convert[0].IsPublic.ShouldBeFalse();
+    }
```
jakhog: Also check that it calles the caller correctly.

Specifications/Events/for_EventStore/when_committing_an_object.cs
```diff
@@ -0,0 +1,33 @@
...
+            event_types.Setup(_ => _.GetFor(content.GetType())).Returns(event_type);
+        };
+
+        Because of = async () => await event_store.Commit(content, event_source);
+        It should_convert_the_object_to_uncommitted_events = () => events_to_convert.ShouldBeOfExactType<UncommittedEvents>();
+        It should_have_the_correct_event_source_id = () => events_to_convert[0].EventSource.ShouldEqual(event_source);
+        It should_have_the_correct_event_type = () => events_to_convert[0].EventType.ShouldEqual(event_type);
+        It should_have_the_correct_content = () => events_to_convert[0].Content.ShouldEqual(content);
+        It shouldnt_be_public = () => events_to_convert[0].IsPublic.ShouldBeFalse();
+    }
```
jakhog: Also check that it calles the caller correctly.

Specifications/Events/for_EventStore/when_committing_a_public_object_with_an_event_type.cs
```diff
@@ -0,0 +1,31 @@
...
+        };
+
+        Because of = async () => await event_store.CommitPublic(content, event_source, event_type);
+        It should_convert_the_object_to_uncommitted_events = () => events_to_convert.ShouldBeOfExactType<UncommittedEvents>();
+        It should_have_the_correct_event_source_id = () => events_to_convert[0].EventSource.ShouldEqual(event_source);
+        It should_have_the_correct_event_type = () => events_to_convert[0].EventType.ShouldEqual(event_type);
+        It should_have_the_correct_content = () => events_to_convert[0].Content.ShouldEqual(content);
+        It should_be_public = () => events_to_convert[0].IsPublic.ShouldBeTrue();
+    }
```
jakhog: Also check that it calles the caller correctly.

Specifications/Events/for_EventStore/when_committing_a_public_object.cs
```diff
@@ -0,0 +1,33 @@
...
+            event_types.Setup(_ => _.GetFor(content.GetType())).Returns(event_type);
+        };
+
+        Because of = async () => await event_store.CommitPublic(content, event_source);
+        It should_convert_the_object_to_uncommitted_events = () => events_to_convert.ShouldBeOfExactType<UncommittedEvents>();
+        It should_have_the_correct_event_source_id = () => events_to_convert[0].EventSource.ShouldEqual(event_source);
+        It should_have_the_correct_event_type = () => events_to_convert[0].EventType.ShouldEqual(event_type);
+        It should_have_the_correct_content = () => events_to_convert[0].Content.ShouldEqual(content);
+        It should_be_public = () => events_to_convert[0].IsPublic.ShouldBeTrue();
+    }
```
jakhog: Also check that it calles the caller correctly.

jakhog (changes requested):

Specifications/Events/for_EventStore/given/an_event_store_and_an_execution_context.cs
joelhoisko: 1. The setup on the `converter.ToProtobuf` with the callbacks to access the events felt a bit hacky for me so I think that the comment provides some context to it as I don't think it's self evident what I was doing. The upper comment was more for me in the moment that I just left in so I agree on that one.
2. I'm also a bit surprised how everything seemed to work regarding the `.Returns()` as thing's we're also quite happy with accepting null or even outright wrong types here too so that probably kinda drove me deeper into it.
3. What do you mean by returning a specific object (`CommitEventsResult/UncommittedEvents`)?
As in having the converter not be a mock and working correctly (which means that we'd have to setup the `Call()` to have a whole bunch of logic to return a proper `Contracts.CommittedEvent` from the given `Contracts.UncommittedEvents` like the real `EventStoreClient` would do) or the converter would just return an empty `CommitEventsResult/UncommittedEvents`?

joelhoisko:
Did some changes
- Doesn't track the `UncommittedEvents` passed to converter anymore, just checks that the `converter.ToProtobuf()` is always called with `UncommittedEvents`.
- `converter.ToProtobuf()` returns a reference to an `IEnumerable<Contracts.UncommittedEvent>`.
-  Checks that the `caller.Call()` is always called with the referenced `Contracts.CommitEventsRequest`.
- Make `caller.Call()` return a reference to `Contracts.CommitEventsResponse`.
- Check that `converter.ToSDK()` is always called with the referenced `Contracts.CommitEventsResponse`.
- Check that `caller.Call()` is only called once.
- Check that with an unknown event type that the `EventStore` crashes properly and doesn't call the converter/caller etc.

Specifications/Events/for_EventStore/when_committing_a_public_event_with_an_event_type.cs
```diff
@@ -0,0 +1,34 @@
...
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye public world", 12345, true);
+            event_source = new EventSourceId("be274a66-e442-455a-b5d9-81d058fc179e");
+            event_type = new EventType("31819551-844f-4a6f-935c-917a126531f6");
+        };
+
+        Because of = async () => await event_store.CommitPublic(content, event_source, event_type);
```
jakhog: As far as I know - `async` is not supported by Machine.Specifications. So we need to wait with a blocking call.

I think we also want to make sure that the returned result from the `CommitPublic` is the result from the `EventConverter` - right?

Specifications/Events/for_EventStore/when_committing_a_public_event_with_an_event_type.cs
```diff
@@ -0,0 +1,34 @@
...
+        It should_not_call_the_event_types_with_the_content = () => event_types.Verify(_ => _.GetFor(content.GetType()), Times.Never());
+        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()));
```
jakhog, line 29: I would think that we want to verify that the uncommitted event sent in here contains the `content`, `event_source` and `event_type` that is created here at the top?

Specifications/Events/for_EventStore/when_committing_a_public_event_with_known_type.cs
```diff
@@ -0,0 +1,36 @@
...
+        Establish context = () =>
+        {
+            content = new an_event("goodbye public world", 12345, true);
+            event_source = new EventSourceId("9abbd75b-9d7e-401b-92a4-e541233da537");
+            event_type = new EventType("adbdc0df-c4f1-42af-be42-695ba4d6c1fc");
+
+            event_types.Setup(_ => _.GetFor(content.GetType())).Returns(event_type);
+        };
+
+        Because of = async () => await event_store.CommitPublic(content, event_source);
```
jakhog: Same here - async and check result.

Specifications/Events/for_EventStore/when_committing_a_public_event_with_known_type.cs
```diff
@@ -0,0 +1,36 @@
...
+        It should_call_the_event_types_with_the_content = () => event_types.Verify(_ => _.GetFor(content.GetType()));
+        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()));
```
jakhog, line 31: Same here, check the uncommitted event passed to the converter.

Specifications/Events/for_EventStore/when_committing_an_event.cs
```diff
@@ -0,0 +1,37 @@
...
+        Establish context = () =>
+        {
+            content = new an_event("goodbye world", 12345, true);
+            event_source = new EventSourceId("e4799653-eb7d-481b-9548-3156ddb45832");
+            event_type = new EventType("843c431e-7cd6-43d0-a3e2-e82d6da0218b");
+
+            event_types.Setup(_ => _.GetFor(content.GetType())).Returns(event_type);
+        };
+
+        Because of = async () => await event_store.Commit(content, event_source);
```
jakhog: Async + result

Specifications/Events/for_EventStore/when_committing_an_event.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        It should_call_the_event_types_with_the_content = () => event_types.Verify(_ => _.GetFor(content.GetType()));
+        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()));
```
jakhog, line 32: Check the uncommitted events

Specifications/Events/for_EventStore/when_committing_an_event_with_an_unknown_event_type.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            try
+            {
+                await event_store.Commit(content, event_source);
+            }
+            catch (Exception e)
+            {
+                exception = e;
+            }
+        };
```
jakhog: No async, and you can use `exception = Catch.Exception(...)`, there should be examples of that in the other specifications somewhere.

Specifications/Events/for_EventStore/when_committing_an_event_with_event_type.cs
```diff
@@ -0,0 +1,34 @@
...
+        static EventType event_type;
+
+        Establish context = () =>
+        {
+            content = new an_event("goodbye world", 12345, false);
+            event_source = new EventSourceId("6bb48542-29b7-4339-a711-b8c8ca733f9d");
+            event_type = new EventType("b40a6161-44dd-481f-a1ff-40694c05f91d");
+        };
+
+        Because of = async () => await event_store.Commit(content, event_source, event_type);
```
jakhog: Async + result

Specifications/Events/for_EventStore/when_committing_an_event_with_event_type.cs
```diff
@@ -0,0 +1,34 @@
...
+        It should_not_call_the_event_types_with_the_content = () => event_types.Verify(_ => _.GetFor(content.GetType()), Times.Never());
+        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()));
```
jakhog, line 29: Check the uncommitted events

Specifications/Events/for_EventStore/when_committing_an_uncommitted_event.cs
```diff
@@ -19,17 +21,18 @@ public class when_committing_an_uncommitted_event : given.an_event_store_and_an_
...
             content = new an_event("goodbye world", 12345, false);
-            event_source = new EventSourceId(Guid.NewGuid());
-            event_type = new EventType(Guid.NewGuid());
+            event_source = new EventSourceId("cc82c47c-2012-4916-9452-1215c96a5550");
+            event_type = new EventType("18e54777-fddb-4d48-b86d-05161c3dc288");
             is_public = false;
             uncommitted_event = new UncommittedEvent(event_source, event_type, content, is_public);
         };
 
         Because of = async () => await event_store.Commit(uncommitted_event);
```
jakhog: Async + result

Specifications/Events/for_EventStore/when_committing_an_uncommitted_event.cs
```diff
@@ -19,17 +21,18 @@ public class when_committing_an_uncommitted_event : given.an_event_store_and_an_
...
-        It should_convert_the_object_to_uncommitted_events = () => events_to_convert.ShouldBeOfExactType<UncommittedEvents>();
-        It should_have_the_correct_event_source_id = () => events_to_convert[0].EventSource.ShouldEqual(uncommitted_event.EventSource);
-        It should_have_the_correct_event_type = () => events_to_convert[0].EventType.ShouldEqual(uncommitted_event.EventType);
-        It should_have_the_correct_content = () => events_to_convert[0].Content.ShouldEqual(uncommitted_event.Content);
-        It shouldnt_be_public = () => events_to_convert[0].IsPublic.ShouldEqual(uncommitted_event.IsPublic);
+        It should_not_call_the_event_types_with_the_content = () => event_types.Verify(_ => _.GetFor(content.GetType()), Times.Never());
+        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()));
```
jakhog: Check that uncommitted_event is passed along here

Specifications/Events/for_EventStore/when_committing_uncommitted_events.cs
```diff
@@ -39,24 +40,11 @@ public class when_committing_uncommitted_events : given.an_event_store_and_an_ex
         };
 
         Because of = async () => await event_store.Commit(uncommitted_events);
```
jakhog: Async + result

Specifications/Events/for_EventStore/when_committing_uncommitted_events.cs
```diff
@@ -39,24 +40,11 @@ public class when_committing_uncommitted_events : given.an_event_store_and_an_ex
...
-        It should_have_the_same_is_public_for_the_second_event = () => events_to_convert[1].IsPublic.ShouldEqual(uncommitted_events[1].IsPublic);
-        It should_have_the_same_event_source_for_the_third_event = () => events_to_convert[2].EventSource.ShouldEqual(uncommitted_events[2].EventSource);
-        It should_have_the_same_event_type_for_the_third_event = () => events_to_convert[2].EventType.ShouldEqual(uncommitted_events[2].EventType);
-        It should_have_the_same_event_type_generation_for_the_third_event = () => events_to_convert[2].EventType.Generation.ShouldEqual(uncommitted_events[2].EventType.Generation);
-        It should_have_the_same_content_for_the_third_event = () => events_to_convert[2].Content.ShouldEqual(uncommitted_events[2].Content);
-        It should_have_the_same_is_public_for_the_third_event = () => events_to_convert[2].IsPublic.ShouldEqual(uncommitted_events[2].IsPublic);
+
+        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(Moq.It.IsAny<UncommittedEvents>()));
```
jakhog: Check that uncommitted_events is passed along here.
```suggestion
        It should_call_the_converter_with_uncommitted_events = () => converter.Verify(_ => _.ToProtobuf(uncommitted_events));
```

jakhog (changes requested):

Specifications/Events/for_EventStore/when_committing_a_public_event_with_an_event_type.cs
joelhoisko: `async` support seems to be a [_"maybe"_](https://github.com/machine/machine.specifications/issues/293#issuecomment-655909400) so yeah, better to just keep it blocking.

Specifications/Events/for_EventStore/when_committing_an_event_with_an_unknown_event_type.cs
joelhoisko: I tried to use it earlier, but I didn't catch (hah) that `Catch.Exception()` took a function as it's parameter instead of just the call. Fixed it now.

joelhoisko:
- Removed `async`
- Verify that the `UncommittedEvents` passed to `converter.ToProtobuf` contains the same stuff as given. Not sure if this is nicer than saving it in a callback like earlier as the `.Verify()` calls get a bit wieldy
- Check that we get the result from `converer.ToSDK()` from each `Commit` function.

jakhog (approved):

PR #17, opened 2020-09-25
Remove EventIdentifier

jakhog:
Removes the `EventIdentifier` class. We can add it back out if we need it again later, or maybe construct it in the Runtime to make it more consistent.


woksin (approved):
Less code to read 👍

PR #19, opened 2020-09-26
Event handler and Event type class ++

woksin:
Did more stuff here than what I planned. Hope it's not too much 😄 

* Created IContainer interface and a default Container implementation that only can created constructor-less services.
* Added WithContainer(IContainer) builder method to ClientBuilder
* Event handler methods can now return both void and Task. It must return one or the other
* Register Event Handler by just passing the class or the instance of a class. A handle method must either be decorated with HandlesAttribute or have the name Handle.
* Associate Event Type by just passing the class. It throws an exception if class is not decorated with EventType attribute
* Changed the event handler builders so that warnings are aggregated and bubbles up to the top-level builder so that we can log the warnings when building and registering handlers without giving the loggers as constructor arguments.
* Changed the artifacts builder implementation a bit

I'm not that happy with how I implemented the functionality for building event handler from a class. I'm open to suggestions there


woksin:
It seems that I have forgotten to implement the possibility for even handler classes to be decorated with ScopeId

Source/DependencyInversion/Container.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.SDK.DependencyInversion
+{
+    /// <summary>
+    /// Represents an implementation of <see cref="IContainer" />.
+    /// </summary>
+    public class Container : IContainer
```
jakhog:
```suggestion
    public class DefaultContainer : IContainer
```

Source/Events.Handling/Builder/EventHandlersBuilder.cs
```diff
@@ -34,24 +25,70 @@ public EventHandlersBuilder(ILoggerFactory loggerFactory)
...
+        /// Registers a <see cref="Type" /> as an event handler class.
+        /// </summary>
+        /// <param name="eventHandlerInstance">The event handler instance.</param>
+        /// <returns>The <see cref="EventHandlersBuilder" /> for continuation.</returns>
+        public EventHandlersBuilder RegisterEventHandler(object eventHandlerInstance)
+        {
+            _builders.Add(new ConventionEventHandlerBuilder(eventHandlerInstance));
+            return this;
+        }
+
```
jakhog: This seems a bit superfluous, since if you are passing in an object you have created yourself, the compiler will know the type, and should be able to figure out the previous signature.
```suggestion
```

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,311 @@
...
+        /// <inheritdoc/>
+        public BuildEventHandlerResult BuildAndRegister(
+            IEventProcessors eventProcessors,
+            IEventTypes eventTypes,
+            IEventProcessingConverter processingConverter,
+            IContainer container,
+            ILoggerFactory loggerFactory,
+            CancellationToken cancellation)
+        {
+            _eventHandlerInstance ??= container.Get(_eventHandlerType);
```
jakhog: The instance cannot be instantiated here - it needs to be transient and created when an event is handled.

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,311 @@
...
+            var logger = loggerFactory.CreateLogger<ConventionEventHandlerBuilder>();
+            logger.LogTrace("Building event handler from type {EventHandler}", _eventHandlerType);
```
jakhog: Seems like a debug message
```suggestion
            logger.LogDebug("Building event handler from type {EventHandler}", _eventHandlerType);
```

Source/Events.Handling/Builder/BuildEventHandlerResult.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether building the event handler succeeded.
+        /// </summary>
+        public bool Succeeded => Warnings == default;
+
+        /// <summary>
+        /// Gets the <see cref="EventHandlerBuildWarnings" /> - null if no warnings and <see cref="Succeeded" />.
+        /// </summary>
+        public EventHandlerBuildWarnings Warnings { get; }
```
jakhog: This null thing makes my spidey sense tingle. As far as I see this `BuildEventHandlerResult` is a way to capture things that go wrong while building an event handler, to log the warning in one place? It seems to me like it would be easier to just use the logger internally in the builders, and potentially just return a bool?

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        void AddUntypedHandleSignatureToBuilder(MethodInfo method, EventType eventType, EventHandlerMethodsBuilder builder)
+        {
+            var eventHandlerSignatureType = method.ReturnType == typeof(Task) ? typeof(EventHandlerSignature) : typeof(VoidEventHandlerSignature);
+            var eventHandlerSignature = Delegate.CreateDelegate(eventHandlerSignatureType, _eventHandlerInstance, method);
+            var builderHandleMethod = typeof(EventHandlerMethodsBuilder)
+                                        .GetMethod(nameof(EventHandlerMethodsBuilder.Handle), new[] { typeof(EventType), eventHandlerSignatureType });
+            if (eventHandlerSignature is EventHandlerSignature signature) builder.Handle(eventType, signature);
+            else if (eventHandlerSignature is VoidEventHandlerSignature voidSignature) builder.Handle(eventType, voidSignature);
```
jakhog: Seems like there is a chance `Delegate.CreateDelegate` can return null and nothing will be added silently here? Also the if and the first ternary seem to check for the same thing, perhaps we should do the check with a boolean and share that for both of them?

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,319 @@
...
+                                                typeof(TypedEventHandlerSignature<>)
+                                                : typeof(VoidTypedEventHandlerSignature<>);
+            var eventHandlerSignatureType = eventHandlerSignatureGenericTypeDefinition.MakeGenericType(eventParameterType);
+            var eventHandlerSignature = Delegate.CreateDelegate(eventHandlerSignatureType, _eventHandlerInstance, method);
+            var builderHandleMethod = typeof(EventHandlerMethodsBuilder)
+                                        .GetMethods()
+                                        .Where(_ => _.IsGenericMethod && _.Name == nameof(EventHandlerMethodsBuilder.Handle))
+                                        .Single(_ => _.GetParameters()[0].ParameterType.GetGenericTypeDefinition() == eventHandlerSignatureGenericTypeDefinition)
+                                        .MakeGenericMethod(eventParameterType);
+            builderHandleMethod.Invoke(builder, new[] { eventHandlerSignature });
```
jakhog: Wæt is going on here? This might be something that I invented a long time ago, but either way it is not really all that easy to make sense of. Seems like maybe we could make this simpler by making some new implementations of `IEventHandlerMethod` here maybe?

Source/Events.Handling/Builder/EventHandlerBuildWarnings.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        /// <summary>
+        /// Gets the <see cref="IEnumerable{T}" /> of warnings when attempting to build the event handler.
+        /// </summary>
+        public IEnumerable<string> Warnings { get; }
+
+        /// <inheritdoc/>
+        public override string ToString()
+            => $"Failed to build event handler {_eventHandlerId}{System.Environment.NewLine}{string.Join($"{System.Environment.NewLine}\t", Warnings)}";
```
jakhog: Newlines in log messages are not that great if we can avoid them, most systems read things line by line, so if we can avoid them that helps parsing a log. In this example, if you were to filter for warning messages (as this is printed somewhere else), you would se the "Failed to build event handler ...", then not the reasons since they are not warnings (because of newlines).

This is somewhat related to my comment above by using the ILogger directly when warnings occur instead. That would make life a bit easier for devs.

Source/Events.Handling/Builder/EventHandlerMethodsBuilder.cs
```diff
@@ -106,12 +146,13 @@ public EventHandlerMethodsBuilder Handle(EventTypeId eventTypeId, Generation eve
...
-                    _logger.LogWarning("Event handler {EventHandler} already handles event with event type {EventType}", _eventHandlerId, eventType);
-                    HasInvalidMethods = true;
+                    _warnings.Add($"Event handler {_eventHandlerId} already handles event with event type {eventType}");
                 }
             }
 
-            return eventTypesToMethods;
+            var result = new BuildEventHandlerResult(_eventHandlerId, _warnings);
+            if (!result.Succeeded) eventTypesToMethods = null;
+            return result;
```
jakhog: I think we should just do all the checks here at the end, in stead of adding warnings as we go. Since we can't really be sure until we can resolve the eventtypes as you do here. This would simplify this code somewhat.

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
jakhog: Maybe we should make this class without using the `EventHandlerMethodsBuilder`?

Source/Events.Handling/Builder/EventHandlerMethod.cs
```diff
@@ -11,13 +12,25 @@ namespace Dolittle.SDK.Events.Handling.Builder
...
-        readonly EventHandlerSignature _method;
+        readonly Func<object, EventContext, Task> _method;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EventHandlerMethod"/> class.
         /// </summary>
         /// <param name="method">The <see cref="EventHandlerSignature" />.</param>
-        public EventHandlerMethod(EventHandlerSignature method) => _method = method;
+        public EventHandlerMethod(EventHandlerSignature method)
+            => _method = (@event, context) => method(@event, context);
```
jakhog: Why do we need to create a lambda here?

Source/Events.Handling/Builder/EventHandlersBuilder.cs
```diff
@@ -34,24 +25,70 @@ public EventHandlersBuilder(ILoggerFactory loggerFactory)
...
+            IContainer container,
+            ILoggerFactory loggerFactory,
+            CancellationToken cancellation)
         {
+            var logger = loggerFactory.CreateLogger<EventHandlersBuilder>();
             foreach (var builder in _builders)
             {
-                builder.BuildAndRegister(eventProcessors, eventTypes, processingConverter, cancellation);
+                var buildResult = builder.BuildAndRegister(eventProcessors, eventTypes, processingConverter, container, loggerFactory, cancellation);
+                if (!buildResult.Succeeded) logger.LogWarning(buildResult.Warnings.ToString());
```
jakhog: I think it would be easier to just log warnings if something goes wrong during building inside the builders, and if they don't succeed, they will let the dev know that the handler will not be registered. Then we don't actually care about the result here anymore.

Source/Events.Handling/Builder/TypedEventHandlerMethod.cs
```diff
@@ -14,13 +14,25 @@ namespace Dolittle.SDK.Events.Handling.Builder
...
+        readonly Func<T, EventContext, Task> _method;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TypedEventHandlerMethod{T}"/> class.
         /// </summary>
-        /// <param name="method">The <see cref="EventHandlerSignature" />.</param>
-        public TypedEventHandlerMethod(TypedEventHandlerSignature<T> method) => _method = method;
+        /// <param name="method">The <see cref="TypedEventHandlerSignature{T}" />.</param>
+        public TypedEventHandlerMethod(TypedEventHandlerSignature<T> method)
+            => _method = (T @event, EventContext context) => method(@event, context);
```
jakhog: Doesn't this method already match the signature?

Source/Events.Handling/EventHandlerAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+    [AttributeUsage(AttributeTargets.Class)]
+    public class EventHandlerAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EventHandlerAttribute"/> class.
+        /// </summary>
+        /// <param name="eventHandlerId">The unique identifier of the event handler.</param>
+        /// <param name="partitioned">Whether the event handler is partitioned.</param>
+        /// <param name="scopeId">The scope that the event handler handles events in.</param>
+        public EventHandlerAttribute(string eventHandlerId, bool partitioned = true, string scopeId = default)
```
jakhog: How does this end up looking in the code if it is partitioned or scoped, I guess it means one could write
```csharp
[EventHandler("...", partitioned: false, scopeId: "...")].
```
Right?

I don't remember exactly what we agreed here, maybe we are missing something in the sketches. But I think it was _partitioned_ and _inScope_ in TS?

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        bool TryAddDecoratedHandlerMethods(
+            IEnumerable<MethodInfo> methods,
+            EventHandlerId eventHandlerId,
+            EventHandlerMethodsBuilder eventHandlerMethodsBuilder,
+            out EventHandlerBuildWarnings warnings)
+        {
+            warnings = default;
+            var warningMessages = new List<string>();
+            foreach (var method in methods.Where(IsDecoratedHandlerMethod).ToArray())
```
jakhog: No need for `.ToArray()` in a foreach?

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,319 @@
...
+        bool TryAddConventionHandlerMethods(
+            IEnumerable<MethodInfo> methods,
+            EventHandlerId eventHandlerId,
+            IEventTypes eventTypes,
+            EventHandlerMethodsBuilder eventHandlerMethodsBuilder,
+            out EventHandlerBuildWarnings warnings)
+        {
+            warnings = default;
+            var warningMessages = new List<string>();
+            foreach (var method in methods.Where(_ => !IsDecoratedHandlerMethod(_) && _.Name == MethodName).ToArray())
```
jakhog: `.ToArray()`?

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        bool TryBuildHandlerMethods(
+            EventHandlerId eventHandlerId,
+            IEventTypes eventTypes,
+            out IDictionary<EventType, IEventHandlerMethod> eventTypesToMethods,
+            out BuildEventHandlerResult buildResult)
+        {
+            var eventHandlerMethodsBuilder = new EventHandlerMethodsBuilder(eventHandlerId);
+            var publicMethods = _eventHandlerType.GetMethods(BindingFlags.Instance | BindingFlags.DeclaredOnly | BindingFlags.Public);
```
jakhog: I think we used to have warnings if something looked like a Handler method, so had the Handle name or [Handles(..)] attribute now, but is non-public?

Source/Events/EventTypesBuilder.cs
```diff
@@ -13,23 +13,40 @@ namespace Dolittle.SDK.Events
...
+        public override EventTypesBuilder Associate(Type type, EventType eventType)
         {
-            Artifacts.Associate(type, artifact);
+            if (TryGetDecoratedEventType(type, out var decoratedEventType)
+                && decoratedEventType != eventType)
+            {
+                throw new EventTypeDoesNotMatchDecoratedEventType(eventType, decoratedEventType, type);
+            }
+
+            Associations.Add((type, eventType));
```
jakhog, line 38: Why does this sneakily add things into the associations in stead of calling one of the `Associate(...)` methods?

Source/SDK/ClientBuilder.cs
```diff
@@ -178,7 +192,7 @@ public Client Build()
                 Claims.Empty,
                 CultureInfo.InvariantCulture);
 
-            var eventTypes = _eventTypesBuilder.Build();
+            var eventTypes = _eventTypesBuilder.Build(new EventTypes(_loggerFactory.CreateLogger<EventTypes>()));
```
jakhog: It feels a bit strange that it takes in an `EventTypes` and returns an `EventTypes`, it makes it a bit unclear if it modifies the thing you pass in, or if it creates a brand new one. I suggest we rename the method to something like `AddAssociationsInto`, and not don't return anything from it.

jakhog (changes requested):

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
woksin: I don't know if we should do that?

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
woksin: Lazy evaluation. I have burned myself on that before

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
woksin: Lazy evaluation

Source/Events.Handling/EventHandlerAttribute.cs
woksin: That would be correct, it would look like you wrote there.

Yes, you are correct. I'll change it to inScope

woksin:
I have done a couple of commits that should resolve most of your comments @jakhog. Please have another look :)

Source/SDK/ClientBuilder.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/TypedEventHandlerMethod.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/EventHandlerAttribute.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/EventHandlersBuilder.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/EventHandlerMethod.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/EventHandlerMethodsBuilder.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/EventHandlerBuildWarnings.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/BuildEventHandlerResult.cs
woksin: This should be resolved by latest commits

Source/Events.Handling/Builder/ConventionEventHandlerBuilder.cs
woksin: This should be resolved by latest commits

jakhog (approved):

PR #22, opened 2020-09-30
Resilience

jakhog:
Adds RetryPolicy, and extension methods for retrying and observable with a policy and a cancellation token.

Also uses the RetryPolicy (provided from the ClientBuilder) for registrations of event processors.


woksin (approved):
I like it

PR #24, opened 2020-10-01
Make public APIs adhere to the specification

jakhog:
This PR mostly renames methods and parameters to align with what we have decided in the SDK specification.

Additionally while updating the EventTypesBuilder, I ended up changing the artifacts system a bit, more specifically:
- Removed ArtifactsBuilderFor, IArtifactsBuilderFor, and IArtifact as they provided little reusable code, and resulted in strange API documentation (that mentioned artifact everywhere)
- Changed the Artifacts so that the implementors create their own exceptions, again to not expose the artifact name outside.

I also followed a lot of the suggestions that OmniSharp gave me for a little bit of cleaner code.


Source/Events/EventTypesBuilder.cs
```diff
@@ -2,51 +2,152 @@
...
-                throw new EventTypeDoesNotMatchDecoratedEventType(eventType, decoratedEventType, type);
+                eventTypes.Associate(type, eventType);
             }
+        }
 
-            Associations.Add((type, eventType));
-            return this;
+        void AddAssociation(Type type, EventType eventType)
+        {
+            ThrowIfEventTypeFromAttributeDoesNotMatchProvidedEventType(type, eventType);
```
joelhoisko:
```suggestion
            ThrowIfAttributeSpecifiesADifferentEventType(type, eventType);
```

Source/Events/EventTypesBuilder.cs
```diff
@@ -2,51 +2,152 @@
...
-            => Associate(type, new EventType(artifactId, generation));
+        void ThrowIfTypeIsMissingEventTypeAttribute(Type type)
+        {
+            if (!TryGetEventTypeFromAttribute(type, out _))
+            {
+                throw new TypeIsMissingEventTypeAttribute(type);
+            }
+        }
+
+        void ThrowIfEventTypeFromAttributeDoesNotMatchProvidedEventType(Type type, EventType providedType)
```
joelhoisko:
```suggestion
        void ThrowIfAttributeSpecifiesADifferentEventType(Type type, EventType providedType)
```

joelhoisko (changes requested):

joelhoisko (approved):

PR #34, opened 2020-10-13
Fix AggregateRootVersion in CommittedAggregateEvents.

jakhog:
The AggregateRootVersion on CommittedAggregateEvents was set to one less than it should be.

The version on the composed object (CommittedAggregateEvents) should be the same as the on in the aggregates collection in mongo, which is always equal to the number of evens applied by the aggregate. For each of the committed events, we also store the version of the aggregate that applied the event, which corresponds to the version __before__ the event was committed. The current implementation therefore returned a number that was off by one, and is fixed by this new code.


jakhog:
Nevermind, this was actually wrong.

PR #36, opened 2020-10-17
Take a stab at aggregates

woksin:
## Summary

Adds aggregates to the SDK in a barebones matter, agnostic to whether there is IoC or not.

### Added

- System for working with aggregates and aggregate roots
- AggregateOf<> method on Client for creating an aggregate root operation on a specific aggregate
- Small sample showcasing the aggregates system
- Specs around AggregateRoot

### Changed
- docker-compose in samples run with latest runtime versions


einari (approved):
This is a much needed and yearned for feature :) 

Great to see it finally come here. 
API surface is familiar and looks good.

Issue #39, opened 2020-10-23
Register an instance of IEventTypes


woksin:
Instead of building the event types on the client with the builder we could have an additional method that takes in an instance of IEventTypes

PR #40, opened 2020-10-23
With event types instance

woksin:
Implements feature #39


einari:
Is this still a valid PR ?

woksin:
I remember I wanted it for something, but I rhink we can disregard this

jakhog:
If this is something we need, we can redo it :)

PR #42, opened 2020-10-27
Fix calling Un/Partitioned() multiple times in EventHandlerBuilder 

joelhoisko:
It was possible to write client building code like:
```csharp
_.CreateEventHandle("some-guid", builder =>
{
    // on the inside, calling Partitioned() sets builder._methodsBuilder = new EventHandlerMethodsBuilder()
    builder.Partitioned()
        // this will never be called
        .Handle<MyEvent>((@event, ctx));
    // Partitioned() once again does builder._methodsBuilder = new EventHandlerMethodsBuilder()
    // this removes all of the old handle methods
    builder.Partitioned();
        .Handle<MyOtherEvent>((@event, ctx));
})
```
and as you we're working with the callback it wasn't obvious that this was a trap.

This fixes this by providing a callbackless API:
```csharp
eventHandlersBuilder.CreateEventHandler("44ac0f75-9f13-45f1-b3b1-3f222f92ca57")
    .Partitioned()
    .Handle<MyEvent>((@event, ctx) => Console.WriteLine("A second handle method for MyEvent."))
    .Handle<MyOtherEvent>((@event, ctx) => Console.WriteLine("A handle method for MyOtherevent."));

```

### DIFF
--- a/Samples/Advanced/MyOtherEvent.cs
+++ b/Samples/Advanced/MyOtherEvent.cs
@@ -1,0 +1,19 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Dolittle.SDK.Events;
+
+namespace Basic
+{
+    [EventType("2fc3df44-0ee6-41d7-b0eb-eb2b841894df")]
+    class MyOtherEvent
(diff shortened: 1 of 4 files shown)


woksin (approved):
LGTM

PR #43, opened 2020-10-28
8.0.0 redpill

woksin:
* Changes the builder APIs
* Commit to Aggregate
* Fetch from Aggregate
* Fix for IoC
* Wait removed from Client
* Start added to Client
* Need to call Start() on Client to register the event processors


woksin:
Replaces #32

PR #56, opened 2021-04-07
Get projections

woksin:
## Summary

Adds the ability to get projections.

### Added

- API for getting one or all projections of a specific projection type


Source/Projections/Builder/ProjectionBuilder.cs
```diff
@@ -49,9 +53,11 @@ public ProjectionMethodsForReadModelBuilder<TReadModel> ForReadModel<TReadModel>
         {
             if (_methodsBuilder != default)
             {
-                // Throw exceptionee c
+                var genericType = _methodsBuilder.GetType().GetGenericArguments()[0];
+                throw new ReadModelAlreadyDefinedForProjection(_projectionId, _scopeId, genericType);
```
jakhog: This feels very sneaky. I remember we also kept the type as a property in this class in TS - feels like a valuable thing to do here as well to make this a bit easier on the eyes :)

Source/Projections/Builder/ProjectionsBuilder.cs
```diff
@@ -18,6 +19,18 @@ namespace Dolittle.SDK.Projections.Builder
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ProjectionsBuilder"/> class.
+        /// </summary>
+        /// <param name="projectionAssociations">The <see cref="IProjectionAssociations"/>.</param>
+        public ProjectionsBuilder(IProjectionAssociations projectionAssociations)
+        {
+            _projectionAssociations = projectionAssociations;
+
+            Console.WriteLine($"Projections builder associations hash {_projectionAssociations.GetHashCode()}");
```
jakhog: Probably we don't want this :p
```suggestion
```

Source/Projections/Store/Builders/ProjectionsBuilder.cs
```diff
@@ -0,0 +1,71 @@
...
+using Dolittle.SDK.Tenancy;
+using Microsoft.Extensions.Logging;
+using ExecutionContext = Dolittle.SDK.Execution.ExecutionContext;
+
+namespace Dolittle.SDK.Projections.Store.Builders
+{
+    /// <summary>
+    /// Represents a builder for building <see cref="Projections"/>.
+    /// </summary>
+    public class ProjectionsBuilder
```
jakhog: Let's rename this so we have less collisions in the Client etc.
```suggestion
    public class ProjectionStoreBuilder
```

Source/Projections/Store/Builders/ProjectionsBuilder.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Dolittle.SDK.Execution;
+using Dolittle.SDK.Projections.Store.Converters;
+using Dolittle.SDK.Services;
+using Dolittle.SDK.Tenancy;
+using Microsoft.Extensions.Logging;
+using ExecutionContext = Dolittle.SDK.Execution.ExecutionContext;
```
jakhog: Whats going on here? Seems it should not be there...
```suggestion
```

Source/Projections/Store/Converters/CouldNotDeserializeProjection.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class CouldNotDeserializeProjection : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CouldNotDeserializeProjection"/> class.
+        /// </summary>
+        /// <param name="state">The current projection state.</param>
+        /// <param name="type">The <see cref="ProjectionCurrentStateType" />.</param>
+        /// <param name="innerException">The inner deserialization <see cref="Exception" />.</param>
+        public CouldNotDeserializeProjection(string state, ProjectionCurrentStateType type, Exception innerException)
+            : base($"Could ont deserialize projection with current state type {type}. State is {state}", innerException)
```
jakhog:
```suggestion
            : base($"Could not deserialize projection with state type {type} and state '{state}'", innerException)
```

Source/Projections/Store/Converters/IConvertProjectionsToSDK.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// Convert from <see cref="IEnumerable{T}"/> of type <see cref="PbCurrentState"/> to <see cref="IEnumerable{T}" /> of <see cref="CurrentState{TProjection}"/>.
+        /// </summary>
+        /// <param name="source"><see cref="IEnumerable{T}"/> of type <see cref="PbCurrentState"/>.</param>
+        /// <param name="states">When the method returns, the converted <see cref="IEnumerable{T}" /> of <see cref="CurrentState{TProjection}"/> if conversion was successful, otherwise null.</param>
+        /// <param name="error">When the method returns, null if the conversion was successful, otherwise the error that caused the failure.</param>
+        /// <typeparam name="TProjection">The <see cref="Type" /> of the projection.</typeparam>
+        /// <returns>A value indicating whether or not the conversion was successful.</returns>
+        bool TryConvert<TProjection>(IReadOnlyList<PbCurrentState> source, out IEnumerable<CurrentState<TProjection>> states, out Exception error)
```
jakhog: Should be able to use this right?
```suggestion
        bool TryConvert<TProjection>(IEnumerable<PbCurrentState> source, out IEnumerable<CurrentState<TProjection>> states, out Exception error)
```

Source/Projections/Store/CurrentState.cs
```diff
@@ -0,0 +1,34 @@
...
+        public CurrentState(TProjection state, CurrentStateType type)
+        {
+            State = state;
+            Type = type;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="CurrentStateType" />.
+        /// </summary>
+        public CurrentStateType Type { get; }
```
jakhog: I'm wondering if we should expose the state type as what it means instead of the enum. Like in `ProjectionContext` - this is turned into a property `WasCreatedFromInitialState`. Feels to me like it makes it more clear to the consumer.

Source/Projections/Store/CurrentState.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Gets the state.
+        /// </summary>
+        public TProjection State { get; }
```
jakhog, line 36: Maybe we should also add an implicit cast operator to `TProjection` so that it's easy to pass around if you only really care about the data?

Source/Projections/Store/CurrentStateType.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Dolittle.SDK.Projections.Store
+{
+    /// <summary>
+    /// Defines the different types a projection state can have.
+    /// </summary>
+    public enum CurrentStateType
```
jakhog, line 9: We should probably also use this in the `ProjectionContext` somehow - that code is slightly hacky right now. I think it is located in the event processor for a projection I think.

Source/Projections/Store/IProjections.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Dolittle.SDK.Projections.Store
+{
+    /// <summary>
+    /// Defines an interface for working directly with Projections.
+    /// </summary>
+    public interface IProjections
```
jakhog: Should we also rename this to be consistent?
```suggestion
    public interface IProjectionStore
```

Source/Projections/Store/IProjectionAssociations.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System;
+using Dolittle.SDK.Events;
+
+namespace Dolittle.SDK.Projections.Store
+{
+    /// <summary>
+    /// Defines a associations for projections.
+    /// </summary>
+    public interface IProjectionAssociations
```
jakhog: In the spirit of naming, maybe:
```suggestion
    public interface IProjectionReadModelTypeAssociations
```

Source/Projections/Store/NoTypeAssociatedWithProjection.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using Dolittle.SDK.Events;
+
+namespace Dolittle.SDK.Projections.Store
+{
+    /// <summary>
+    /// Exception that gets thrown when there is no <see cref="Type" /> associated with a projection.
+    /// </summary>
+    public class NoTypeAssociatedWithProjection : Exception
```
jakhog: Why would this happen? I don't see where we would resolve the type from an projection ID right now.

Source/Projections/Store/ProjectionAssociation.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using Dolittle.SDK.Concepts;
+using Dolittle.SDK.Events;
+
+namespace Dolittle.SDK.Projections.Store
+{
+    /// <summary>
+    /// Represents a projection that is associated with a <see cref="Type" />.
+    /// </summary>
+    public class ProjectionAssociation : Value<ProjectionAssociation>
```
jakhog: This sounds more like a `ScopedProjectionIdentifier`

Source/Projections/Store/Projections.cs
```diff
@@ -0,0 +1,119 @@
...
+                ProjectionId = projection.Identifier.ToProtobuf(),
+                ScopeId = projection.ScopeId.ToProtobuf()
+            };
+
+            var response = await _caller.Call(_getOneMethod, request, cancellationToken).ConfigureAwait(false);
+            response.Failure.ThrowIfFailureIsSet();
+
+            if (!_toSDK.TryConvert<TProjection>(response.State, out var state, out var error))
+            {
+                _logger.LogError(error, "The Runtime acknowledges that the projection state was returned, but the returned {State} could not be converted.", response.State);
```
jakhog: Maybe not the right way to do it - but something like:
```suggestion
                _logger.LogError(error, "The Runtime returned the projection state '{State}'. But it could not be converted to {typeof(TProjection)}.", response.State);
```

Source/Projections/Store/Projections.cs
```diff
@@ -0,0 +1,119 @@
...
+                ProjectionId = projection.Identifier.ToProtobuf(),
+                ScopeId = projection.ScopeId.ToProtobuf()
+            };
+
+            var response = await _caller.Call(_getAllMethod, request, cancellationToken).ConfigureAwait(false);
+            response.Failure.ThrowIfFailureIsSet();
+
+            if (!_toSDK.TryConvert<TProjection>(response.States, out var states, out var error))
+            {
+                _logger.LogError(error, "The Runtime acknowledges that the projection states was returned, but the returned {States} could not be converted.", response.States);
```
jakhog: Same here as previous comment

jakhog (changes requested):

Source/Projections/Store/Converters/IConvertProjectionsToSDK.cs
woksin: I guess, but we used the IReadOnlyList in the Event store code so this is just a copy pasta of that

Source/Projections/Store/CurrentState.cs
woksin: Yeah that would make sense, let's do that

Source/Projections/Store/CurrentState.cs
woksin: Yeeeah, I like that thinking

Source/Projections/Store/CurrentStateType.cs
woksin: I don't really understand what you want here? To have a property with this type in the ProjectionContext?

Source/Projections/Store/NoTypeAssociatedWithProjection.cs
woksin: I guess that this was put in for consistency. We could remove the possibility of resolving type from id

Source/Projections/Builder/ProjectionBuilder.cs
woksin: A little sneaky. I guess that we can just use the typeof(TReadModel) here I guess

Source/Projections/Store/CurrentStateType.cs
jakhog: Not really. This CurrentStateType feels a bit "internal" to the protocol. And right now, it's just describing if the state you got back for the projection was stored - or if you got the default values. In the `ProjectionContext` we "reduced" this enum into a bool saying `WasCreatedFromInitialState`.

woksin: Ahh. So removing the type?

woksin: As in this type

jakhog: I'll fix this in a later PR. In the projection processor - we skipped the conversion to a Runtime representation of this protobuf enum. So it's just about making it nicer. But I'll do it later :)

woksin: 👍

jakhog (approved):

PR #71, opened 2021-09-17
Adding support for setting default JsonSerializerSettings

### LINKED ISSUE #70: Ability to provide custom JsonSerializerSettings.
In order to provide custom converters and override default settings, it would be great if we could have a way to at the time of building the client to configure `JsonSerializerSettings used by `EventContentSerializer` internally in the Try* serialize / deserialize methods.

```csharp
client
   .ForMicroservice(...)
   .WithJsonSerializerSettings(_ => new JsonSerializerSettings { Converters = .... });
```

einari:
## Summary

Adding the ability to set a default `JsonSerializerSettings` instance for the serialization and deserialization of events. This allows for completely custom settings e.g. adding converters for types or casing configuration or similar. Fixes #70.

Using it would then be as follows during the building of the client:

```csharp
client
   .ForMicroservice(...)
   .WithJsonSerializerSettings(new JsonSerializerSettings { Converters = .... });
```

### Added

- Ability to set a default `JsonSerializerSettings` instance.
- `EventContentSerializer` honoring the default `JsonSerializerSettings`.

### DIFF
--- a/Source/Events/Store/Converters/EventContentSerializer.cs
+++ b/Source/Events/Store/Converters/EventContentSerializer.cs
@@ -19,7 +19,7 @@ namespace Dolittle.SDK.Events.Store.Converters
         /// </summary>
         /// <param name="eventTypes"><see cref="IEventTypes"/> for mapping types and artifacts.</param>
         /// <param name="jsonSerializerSettingsProvider"><see cref="Func{T}"/> that provides <see cref="JsonSerializerSettings"/>.</param>
-        public EventContentSerializer(IEventTypes eventTypes, Func<JsonSerializerSettings> jsonSerializerSettingsProvider = null)
+        public EventContentSerializer(IEventTypes eventTypes, Func<JsonSerializerSettings> jsonSerializerSettingsProvider)
         {
             _eventTypes = eventTypes;
             _jsonSerializerSettingsProvider = jsonSerializerSettingsProvider;
--- a/Source/Events/Store/Converters/EventContentSerializer.cs
+++ b/Source/Events/Store/Converters/EventContentSerializer.cs
@@ -23,7 +23,6 @@ namespace Dolittle.SDK.Events.Store.Converters
         {
             _eventTypes = eventTypes;
             _jsonSerializerSettingsProvider = jsonSerializerSettingsProvider;
-            _jsonSerializerSettingsProvider ??= () => new JsonSerializerSettings();
         }
 
         /// <inheritdoc/>
--- a/Source/Events/Store/Converters/EventContentSerializer.cs
+++ b/Source/Events/Store/Converters/EventContentSerializer.cs
@@ -12,22 +12,18 @@ namespace Dolittle.SDK.Events.Store.Converters
     public class EventContentSerializer : ISerializeEventContent
     {
         readonly IEventTypes _eventTypes;
-        readonly JsonSerializerSettings _jsonSerializerSettings;
-        readonly JsonSerializerExceptionCatcher _exceptionCatcher;
+        readonly Func<JsonSerializerSettings> _jsonSerializerSettingsProvider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EventContentSerializer"/> class.
         /// </summary>
         /// <param name="eventTypes"><see cref="IEventTypes"/> for mapping types and artifacts.</param>
-        /// <param name="jsonSerializerSettings">Optional <see cref="JsonSerializerSettings"/>.</param>
-        public EventContentSerializer(IEventTypes eventTypes, JsonSerializerSettings jsonSerializerSettings = null)
+        /// <param name="jsonSerializerSettingsProvider"><see cref="Func{T}"/> that provides <see cref="JsonSerializerSettings"/>.</param>
+        public EventContentSerializer(IEventTypes eventTypes, Func<JsonSerializerSettings> jsonSerializerSettingsProvider = null)
         {
             _eventTypes = eventTypes;
-            _jsonSerializerSettings = jsonSerializerSettings ?? new JsonSerializerSettings();
-
-            _exceptionCatcher = new JsonSerializerExceptionCatcher();
-            _jsonSerializerSettings.Error = _exceptionCatcher.OnError;
-            _jsonSerializerSettings.Formatting = Formatting.None;
+            _jsonSerializerSettingsProvider = jsonSerializerSettingsProvider;
+            _jsonSerializerSettingsProvider ??= () => new JsonSerializerSettings();
         }
 
         /// <inheritdoc/>
@@ -49,11 +45,16 @@ namespace Dolittle.SDK.Events.Store.Converters
 
         bool TrySerializeWithSettings(object content, out string json, out Exception serializationError)
         {
-            json = JsonConvert.SerializeObject(content, _jsonSerializerSettings);
+            var exceptionCatcher = new JsonSerializerExceptionCatcher();
+            var serializerSettings = _jsonSerializerSettingsProvider();
+            serializerSettings.Error = exceptionCatcher.OnError;
+            serializerSettings.Formatting = Formatting.None;
 
-            if (_exceptionCatcher.Failed)
+            json = JsonConvert.SerializeObject(content, serializerSettings);
+
+            if (exceptionCatcher.Failed)
             {
-                serializationError = new CouldNotSerializeEventContent(content, _exceptionCatcher.Error);
+                serializationError = new CouldNotSerializeEventContent(content, exceptionCatcher.Error);
                 return false;
             }
             else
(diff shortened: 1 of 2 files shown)


einari:
In the issue I suggested using a callback to get the settings instance, something like `Func<JsonSerializerSettings>`. But I simplified this to just take the instance, figured `.WithJsonSerializerSettings(_ => new ...)` wasn't adding much value with the extra lambda syntax.

woksin (changes requested):
Great stuff, @einari! This is a general improvement and a flexibility we should provide. However I would like to point out that events are not the only serialisation object, we also have Projections and Embeddings (that are pretty similar). Since this is an added feature to the API I would us to think about the options we have.

First is whether we only want to provide custom serialisation options for events or do we want to provide the same flexibility for Embeddings and Projections as well. And if so, should it be using the same settings (in that case your `WithJsonSerializerSettings` method makes sense)? For further flexibility we might want to at that point have the option to provide custom settings for each type (`WithEventSerializationSettings`, `WithEmbeddingSerializationSettings`, `WithProjectionSerializationSettings`). If we wanted to go down that road I would suggest just renaming `WithJsonSerializerSettings` to `WithEventSerializationSettings `, and then we can implement the other customisations later.  On the other side, if we only wanted to provide custom serialisation options for events I would still like to rename it to `WithEventSerializationSettings`.

I don't have the answers here, I think this is something we need to figure out together with @jakhog and @joelhoisko 

Other than that your implementation is sound :)

einari:
Good point. I was very focused on just the basic events right now. But any events being serialized with for instance a Converter to handle a property with a certain specific type would need to be handled the same way either its Commit, event handler, embedding or projection - otherwise that would in worst case scenario throw an exception when deserializing.

There could be the case for custom serialization settings for the different things you're doing, so one could argue for this on multiple levels, say for instance a general configuration which would become the default one and then override it for each scenario.

Source/Events/Store/Converters/EventContentSerializer.cs
```diff
@@ -12,14 +12,22 @@ namespace Dolittle.SDK.Events.Store.Converters
...
         /// </summary>
         /// <param name="eventTypes"><see cref="IEventTypes"/> for mapping types and artifacts.</param>
-        public EventContentSerializer(IEventTypes eventTypes)
+        /// <param name="jsonSerializerSettings">Optional <see cref="JsonSerializerSettings"/>.</param>
+        public EventContentSerializer(IEventTypes eventTypes, JsonSerializerSettings jsonSerializerSettings = null)
         {
             _eventTypes = eventTypes;
+            _jsonSerializerSettings = jsonSerializerSettings ?? new JsonSerializerSettings();
+
+            _exceptionCatcher = new JsonSerializerExceptionCatcher();
```
woksin: @jakhog made the point that we cannot use the `JsonSerializerExceptionCatcher` multiple times like you changed it to do here. Reason being that it will only catch one Exception and hold unto that for the lifetime of the object. A quick fix for this is to create a new instance if `JsonSerializerExceptionCatcher`each time we (de)serialize and set the Error handler to the new exception catcher's OnError

Source/SDK/ClientBuilder.cs
```diff
@@ -238,6 +240,17 @@ public ClientBuilder WithLogging(ILoggerFactory factory)
...
+        /// <summary>
+        /// Sets the default <see cref="JsonSerializerSettings"/> .
+        /// </summary>
+        /// <param name="jsonSerializerSettings">The default <see cref="JsonSerializerSettings"/>.</param>
+        /// <returns>The client builder for continuation.</returns>
+        public ClientBuilder WithJsonSerializerSettings(JsonSerializerSettings jsonSerializerSettings)
+        {
+            _jsonSerializerSettings = jsonSerializerSettings;
+            return this;
+        }
```
woksin: We discussed this internally and landed on that we'd want to provide additional functionality in the future for having custom settings for serializing embeddings and projections as well as events. Since this PR only touches on events we think that it's better to change the name of the method so that it reflects what it really does at the moment
```suggestion
        /// <summary>
        /// Sets the <see cref="JsonSerializerSettings"/> for serializing events.
        /// </summary>
        /// <param name="jsonSerializerSettings">The default <see cref="JsonSerializerSettings"/>.</param>
        /// <returns>The client builder for continuation.</returns>
        public ClientBuilder WithEventSerializerSettings(JsonSerializerSettings jsonSerializerSettings)
        {
            _jsonSerializerSettings = jsonSerializerSettings;
            return this;
        }
```

woksin (changes requested):
Hey, @einari! We've come up with a suggestion on the ClientBuilder method and a request to change the logic in the event serializer due to a bug that I missed in my first review

Source/Events/Store/Converters/EventContentSerializer.cs
einari: I wondered about that one :) Good point.
It does however raise the question of threadability for this in general. If the JsonSerializerSettings is a global object as with this solution and we mutate the one property for handling exceptions on it, there is a chance one will midway through serialization in another thread have side-effects of the same nature. 

Should we rather do a deep clone?

einari:
I've added a question on one of the comments. Once that conversation is concluded, I can do the necessary changes.

Source/Events/Store/Converters/EventContentSerializer.cs
woksin: That is also a good point, @einari. Did not think about that one... You are right, I think that this is not thread-safe. A solution would, as you say to create a deep copy of it every time we (de)serialize, but again that incurs performance cost. I'm trying to look if anyone else has dealt with this scenario before

woksin: Hey, @einari! We've talked about this and we have a suggestion on how to solve this. If we change the `WithEventSerializerSettings` to instead take a callback that takes in a `JsonSerialierSettings` and returns nothing (`Action< JsonSerializerSettings >`) this callback will get a clean instance of `JsonSerializerSettings` and modify it manually. Does that make sense?

einari: Indeed. I can change it to do so.

woksin (changes requested):
Hey again, @einari. What was intended with the comment about making it a callback is actually a little more involved than just having it be a `Func< JsonSerializerSettings >` that returns the `JsonSerializerSettings`, the intention was to make it an `Action<JsonSerializerSettings>`. One of the reasons for doing that is to make it easier for the developer not to shoot himself in the foot by creating an instance on the outside and just returning it in the callback. By making it an `Action<JsonSerializerSettings>` we remove this possibility because we are controlling the input to that callback and then it is the developer's choice to do what he wants with the settings.

However I suggest that the `EventContentSerializer` constructor takes in a `Func<JsonSerializerSettings>` as a parameter, that way we can just call that inside whenever we use it. Then on the outside we give it a function that uses the Action from the `ClientBuilder` to modify a newly instantiated instance of `JsonSerializerSettings` that has the `Error` property set to the `OnError` property of a newly created `JsonSerializerExceptionCatcher`. It's hard to explain with words, but did this make sense? If not then just shout out and I can make a separate PR to your branch with a suggestion :)

einari:
That makes perfect sense :) And very good point. I'll fix right away.

einari:
I was very confident that I understood the task - lets hope the above commit mathces what you had in mind :)

Source/Events/Store/Converters/EventContentSerializer.cs
```diff
@@ -12,14 +12,17 @@ namespace Dolittle.SDK.Events.Store.Converters
...
         readonly IEventTypes _eventTypes;
+        readonly Func<JsonSerializerSettings> _jsonSerializerSettingsProvider;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EventContentSerializer"/> class.
         /// </summary>
         /// <param name="eventTypes"><see cref="IEventTypes"/> for mapping types and artifacts.</param>
-        public EventContentSerializer(IEventTypes eventTypes)
+        /// <param name="jsonSerializerSettingsProvider"><see cref="Func{T}"/> that provides <see cref="JsonSerializerSettings"/>.</param>
+        public EventContentSerializer(IEventTypes eventTypes, Func<JsonSerializerSettings> jsonSerializerSettingsProvider = null)
```
woksin: This cannot have a default value now :)
```suggestion
        public EventContentSerializer(IEventTypes eventTypes, Func<JsonSerializerSettings> jsonSerializerSettingsProvider)
```

Source/SDK/ClientBuilder.cs
```diff
@@ -238,6 +240,17 @@ public ClientBuilder WithLogging(ILoggerFactory factory)
             return this;
         }
 
+        /// <summary>
+        /// Sets the <see cref="JsonSerializerSettings"/> for serializing events.
```
woksin: I suggest a corrected documentation 
```suggestion
        /// Sets a callback that configures the <see cref="JsonSerializerSettings"/> for serializing events.
```

woksin (changes requested):
We're almost there :) Just a couple small things

woksin (approved):
Sry for the late review, @einari. Now this looks good to me and we should be ready to merge this :)

woksin:
@einari This PR (from a fork) uncovered some bugs in the establish-context-action that hindered the release to happen. We need to sort that out at some point, but in the meanwhile I'll make sure this PR gets released

woksin:
Here is the release @einari https://github.com/dolittle/DotNET.SDK/releases/tag/v9.2.0 🥳

einari:
Gr8. Thnx.

PR #73, opened 2021-10-04
Allow EventSourceID and PartitionID to be strings instead of Guids using Contracts v6.0.0

woksin:
## Summary

Implementing the changes introduced by https://github.com/dolittle/Contracts/pull/53. Allowing EventSourceID and PartitionID to be strings, to more easily integrate with events from existing systems.

This is considered a breaking change because it requires a Runtime compatible with Contracts v6 to function.

### Added

- EventSourceID is now a string instead of a Guid.
- PartitionID is now also a string instead of a Guid.

### Fixed

- Aligned names of event type fields throughout messages from Contracts v6.0.0


jakhog (approved):

jakhog:
I suggest we ignore the code climate code repetition issues for now. Not something we have introduced here.

PR #74, opened 2021-10-04
Change eventsource and partition to string

jakhog:
## Summary

Changes the EventSourceId and PartitionId types from Guid to String.

### Added

- New specifications for converters that ensure we still handle EventSources created with Guids properly.

### Changed

- The Contracts version to use the prerelease with string types.
- EventSourceId and PartitionId concepts from Guid to String.
- The SDK <-> Protobuf converters as necessary.
- The default value for a PartitionId from "00000000-0000-0000-0000-000000000000" to "".
- Existing specifications now use strings that doesn't look like Guids.

### Removed

- Specifications that tested behaviour when a _null_ EventSourceId was received over gRPC (this is not allowed in Protobuf).


Source/Events/PartitionId.cs
```diff
@@ -9,25 +9,25 @@ namespace Dolittle.SDK.Events
...
     /// Represents the concept of a unique identifier for a partition.
     /// </summary>
-    public class PartitionId : ConceptAs<Guid>
+    public class PartitionId : ConceptAs<string>
     {
         /// <summary>
         /// The <see cref="PartitionId"/> that refers to no partition.
         /// </summary>
-        public static readonly PartitionId Unspecified = Guid.Empty;
+        public static readonly PartitionId Unspecified = string.Empty;
```
woksin: I was unsure of this when I worked on the JavaScript part of this. I think what I did is that I set Unspecified to be the empty guid, as it was before. What do we actually want? We may have some documentation that documents that

Source/Events/PartitionId.cs
```diff
@@ -9,25 +9,25 @@ namespace Dolittle.SDK.Events
...
 
         /// <summary>
         /// Implicitly converts from a <see cref="Guid"/> to an <see cref="PartitionId"/>.
         /// </summary>
         /// <param name="partitionId">The <see cref="Guid"/> representation.</param>
         /// <returns>The converted <see cref="PartitionId"/>.</returns>
-        public static implicit operator PartitionId(Guid partitionId) => new PartitionId { Value = partitionId };
+        public static implicit operator PartitionId(Guid partitionId) => new PartitionId { Value = partitionId.ToString() };
```
woksin, line 24: In JavaScript I removed all the Guid parameters from the `from` static method on the partition and event source concepts, to not cause confusion and to break the API so that we're forced to give strings rather that Guids. I think that is what we want, what do you think?

Source/Events/EventSourceId.cs
```diff
@@ -9,24 +9,24 @@ namespace Dolittle.SDK.Events
...
     /// </summary>
-    public class EventSourceId : ConceptAs<Guid>
+    public class EventSourceId : ConceptAs<string>
     {
         /// <summary>
         /// Implicitly convert from a <see cref="Guid"/> to an <see cref="EventSourceId"/>.
         /// </summary>
         /// <param name="eventSourceId">EventSourceId as <see cref="Guid"/>.</param>
-        public static implicit operator EventSourceId(Guid eventSourceId) => new EventSourceId { Value = eventSourceId };
+        public static implicit operator EventSourceId(Guid eventSourceId) => new EventSourceId { Value = eventSourceId.ToString() };
```
woksin, line 18: Same as I mentioned for PartitionId

woksin (changes requested):
I have some small comments/questions related to the concept classes before we can pull this in

Source/Events/PartitionId.cs
jakhog: Let's leave this for now until we get the results back from the poll.

Source/Events/EventSourceId.cs
jakhog: Let's leave this for now until we get the results back from the poll.

woksin (approved):
As you said we'll see the results from the poll

PR #77, opened 2021-10-25
Release Codename 'Aragorn'

woksin:
## Summary

Fixes a problem with aggregates so that it now can be used in an async way. Registration of event types and aggregate roots to the Runtime.

### Added

- Registration of alias for event types through the attribute or builder
- Registration of alias fro aggregate roots through the attribute
- Default alias registration for event type classes and aggregate root classes. Default is the name of the class
- Extension Methods on the ClientBuilder for registering all event handlers, event types, projections, embeddings, aggregate roots by discovering them through the assemblies and DLLs in the solution 

### Changed

- Log Level for registered Event Processors to Information
- The Getting Started, Aggregates, Projections and Embeddings tutorials updated to reflect new features of the SDK

### Fixed

- A problem where you couldn't use an aggregate root in an asynchronous manner


Samples/Tutorials/GettingStarted/Program.cs
```diff
@@ -15,8 +15,7 @@ public static void Main()
                 .ForMicroservice("f39b1f61-d360-4675-b859-53c05c87c0e6")
                 .WithEventTypes(eventTypes =>
                     eventTypes.Register<DishPrepared>())
-                .WithEventHandlers(builder =>
-                    builder.RegisterEventHandler<DishHandler>())
+                .WithEventHandlers()
```
jakhog: IMO we should present the explicit registration by type as the default until we get familiar with how the assembly loading works. Also I like when the TS/C# code are as similar as possible...

jakhog: Also seeing that there is now another sample specifically for this, then I guess we don't need it here :)

Source/Aggregates/AggregateRootType.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using Dolittle.SDK.Artifacts;
+using Dolittle.SDK.Events;
+
+namespace Dolittle.SDK.Aggregates
+{
+    /// <summary>
+    /// Represents the type of an event.
+    /// </summary>
+    public class AggregateRootType : Artifact<AggregateRootId>
```
jakhog, line 12: In the Runtime we now called this thing an "AggregateRoot", and then we used "AggregateRootInstance" for what is called "AggregateRoot". A little inconsistent - but I guess we'll have to live with that for now.

Source/Aggregates/Builders/AggregateRootsBuilder.cs
```diff
@@ -0,0 +1,118 @@
...
+                return true;
+            }
+
+            aggregateRootType = default;
+            return false;
+        }
+
+        void AddAssociation(Type type, AggregateRootType aggregateRootType)
+        {
+            ThrowIfAttributeSpecifiesADifferentAggregateRootType(type, aggregateRootType);
```
jakhog: We don't need this code here? I assume it is just a result of copy-pasting from the Event Types, where we allow users to send in both arguments with `Associate( , )`?

Source/Aggregates/Builders/ProvidedAggregateRootTypeDoesNotMatchAggregateRootTypeFromAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System;
+using Dolittle.SDK.Events;
+
+namespace Dolittle.SDK.Aggregates.Builders
+{
+    /// <summary>
+    /// Exception that gets thrown when an <see cref="AggregateRootType" /> is being associated to a <see cref="Type" /> that has an <see cref="AggregateRootAttribute" /> that specifies another <see cref="AggregateRootType"/>.
+    /// </summary>
+    public class ProvidedAggregateRootTypeDoesNotMatchAggregateRootTypeFromAttribute : Exception
```
jakhog: I think this is unnecessary, see previous comment...

Source/Aggregates/CannotAssociateMultipleAggregateRootTypesWithType.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Dolittle.SDK.Aggregates
+{
+    /// <summary>
+    /// Exception that gets thrown when attempting to associate multiple instance of <see cref="AggregateRootType"/> with a single <see cref="Type"/>.
+    /// </summary>
+    public class CannotAssociateMultipleAggregateRootTypesWithType : Exception
```
jakhog, line 11: This should never happen, but I guess we can keep it to make it easier to use the `Artifacts< , >` thing.

Source/Events/EventType.cs
```diff
@@ -13,31 +13,64 @@ public class EventType : Artifact<EventTypeId>
...
-        void ThrowIfEventTypeIdIsNull(EventTypeId id)
+        /// <summary>
+        /// Gets the alias for the Event Type.
+        /// </summary>
+        public EventTypeAlias Alias { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the Event Type has an alias or not.
+        /// </summary>
+        public bool HasAlias => Alias?.Value != default;
```
jakhog: Always makes me uneasy when we have to check nulls, is it not possible to set this boolean somehow from the constructors?

Source/SDK/ClientBuilder.cs
```diff
@@ -158,6 +165,25 @@ public ClientBuilder WithCancellation(CancellationToken cancellation)
...
+                asm =>
+                {
+                    _embeddingsBuilder.RegisterAllFrom(asm);
+                    _projectionsBuilder.RegisterAllFrom(asm);
+                    _aggregateRootsBuilder.RegisterAllFrom(asm);
+                    _eventHandlersBuilder.RegisterAllFrom(asm);
+                    _eventTypesBuilder.RegisterAllFrom(asm);
+                },
+                assemblyFilter,
+                includeExeFiles);
```
jakhog: Oh, this is the fancy thing. Let's have a chat about the API here. It looks a little bit scary.

Source/SDK/ClientBuilder.cs
```diff
@@ -364,5 +448,27 @@ async Task EventHorizonRetryPolicy(Subscription subscription, ILogger logger, Fu
...
+            foreach (var assembly in GetAssemblies(assemblyFilter, includeExeFiles))
+            {
+                doWithAssembly(assembly);
+            }
+
+            return this;
+        }
+
+        Task RegisterEventTypes(MethodCaller methodCaller, ExecutionContext executionContext, EventTypes eventTypes)
+            => new Events.Internal.EventTypes(methodCaller, executionContext, _loggerFactory.CreateLogger<Events.Internal.EventTypes>()).Register(eventTypes, _cancellation);
```
jakhog: Can we make a client for this thing as well like the aggregate roots thing - I think that looks nicer (and then they are the same) :)

Source/SDK/ClientBuilder.cs
```diff
@@ -364,5 +448,27 @@ async Task EventHorizonRetryPolicy(Subscription subscription, ILogger logger, Fu
...
+
+        IEnumerable<Assembly> GetAssemblies(Func<Assembly, bool> assemblyFilter, bool includeExeFiles)
+        {
+            var logger = _loggerFactory.CreateLogger<ClientBuilder>();
+            return AssemblyFinder.FindAssemblies(
+                failedFile => logger.LogWarning("Failed to load assembly from file {File}", failedFile),
+                assemblyFilter ?? (_ => true),
+                includeExeFiles);
+        }
```
jakhog: A few Assembly loading things ended up in here now - maybe we should move them out to another thing?

jakhog (approved):
LGTM

PR #98, opened 2021-11-17
Fix resources api and update mongodb driver version

woksin:
## Summary

Fixes the resources API and updates the mongodb driver version


Source/Resources/Resources.csproj
```diff
@@ -15,6 +15,7 @@
     <ProjectReference Include="../Concepts/Concepts.csproj" />
     <ProjectReference Include="../Services/Services.csproj" />
     <ProjectReference Include="../Execution/Execution.csproj" />
+    <ProjectReference Include="..\Async\Async.csproj" />
```
jakhog: I like when all the slashes are the same way :) I can fix this - and some newlines a few places.

versions.props
```diff
@@ -6,7 +6,7 @@
         <RxVersion>4.4.1</RxVersion>
         <ProtobufVersion>3.18.1</ProtobufVersion>
         <GrpcVersion>2.41.0</GrpcVersion>
-        <MongoDBDriverVersion>2.10.1</MongoDBDriverVersion>
+        <MongoDBDriverVersion>2.13.2</MongoDBDriverVersion>
```
jakhog, line 9: Just need to check to be sure - this is compatible with the MongoDB we provide in the platform right?

jakhog (changes requested):
Looks good - I'll just add some newlines and fix some slashes :)

Source/Resources/Resources.csproj
jakhog: Done

versions.props
jakhog: It is :)

jakhog (approved):

Source/Resources/MongoDB/IMongoDBResource.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public interface IMongoDBResource : IResource
+    {
+        /// <summary>
+        /// Gets the <see cref="IMongoDatabase"/> that is configured for this <see cref="IResource"/>.
+        /// </summary>
+        /// <param name="databaseSettingsCallback">The <see cref="Action{T}"/> callback for creating <see cref="MongoDatabaseSettings"/> used to create the <see cref="IMongoDatabase"/>.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/>.</param>
+        /// <returns>A <see cref="Task{TResult}"/>that, when resolved, returns the <see cref="IMongoDatabase"/>.</returns>
+        Task<IMongoDatabase> GetDatabase(Action<MongoDatabaseSettings> databaseSettingsCallback = default, CancellationToken cancellationToken = default);
```
jakhog, line 22: Hmm, do we want overloads for this, instead of just the default value? Don't remember how this would look when you want to just provide the cancellation token...

jakhog (changes requested):

PR #116, opened 2022-01-21
Bilbo: Projections GetAll batching

woksin:
## Summary

Uses the new batch streaming method to get all Projection states from the Runtime. This fixes an issue where a large amount of Projection states caused the gRPC client in the SDK to throw an exception because the response message was too big.

### Fixed

- The `IProjectionStore.GetAll` method now uses the new gRPC method that streams results back in batches to fix the issue of too large gRPC messages when a large amount of projection read models have been created.


Source/Projections/Store/Log.cs
```diff
@@ -0,0 +1,27 @@
...
+
+static partial class Log
+{
+    [LoggerMessage(0, LogLevel.Debug, "Getting one state from projection {Projection} of type {ProjectionType} in scope {Scope} with key {Key}")]
+    internal static partial void GettingOneProjection(ILogger logger, Key key, ProjectionId projection, Type projectionType, ScopeId scope);
+
+    [LoggerMessage(0, LogLevel.Debug, "Getting all states from projection {Projection} of type {ProjectionType} in scope {Scope}")]
+    internal static partial void GettingAllProjections(ILogger logger, ProjectionId projection, Type projectionType, ScopeId scope);
+    
+    [LoggerMessage(0, LogLevel.Debug, "Received batch number {BatchNumber} consisting of {NumProjections} projection states")]
```
jakhog:
```suggestion
    [LoggerMessage(0, LogLevel.Trace, "Received batch number {BatchNumber} consisting of {NumProjections} projection states")]
```

Source/Services/ICanCallAServerStreamingMethod.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Google.Protobuf;
+using Grpc.Core;
+
+namespace Dolittle.SDK.Services;
+
+/// <summary>
+/// Defines a duplex streaming method call.
```
jakhog:
```suggestion
/// Defines a server streaming method call.
```

Source/Services/ICanCallAServerStreamingMethod.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+/// <typeparam name="TClientMessage">Type of the <see cref="IMessage">messages</see> that is sent from the client to the server.</typeparam>
```
jakhog:
```suggestion
/// <typeparam name="TClientMessage">Type of the <see cref="IMessage">message</see> that is sent from the client to the server.</typeparam>
```

Source/Services/MethodCaller.cs
```diff
@@ -59,6 +59,27 @@ public MethodCaller(string host, ushort port)
         }
     }
 
+    /// <inheritdoc />
+    public IServerStreamingMethodHandler<TServerMessage> Call<TClientMessage, TServerMessage>(ICanCallAServerStreamingMethod<TClientMessage, TServerMessage> method, TClientMessage request, CancellationToken token)
+        where TClientMessage : IMessage where TServerMessage : IMessage
```
jakhog:
```suggestion
        where TClientMessage : IMessage
        where TServerMessage : IMessage
```

Source/Services/MethodCaller.cs
```diff
@@ -59,6 +59,27 @@ public MethodCaller(string host, ushort port)
...
+                new AsyncServerStreamingCall<TServerMessage>(
+                    new CatchingAsyncStreamReader<TServerMessage>(_host, _port, originalStream.ResponseStream),
+                        originalStream.ResponseHeadersAsync,
+                        originalStream.GetStatus,
+                        originalStream.GetTrailers,
+                        originalStream.Dispose));
+        }
+        catch (Exception)
+        {
+            throw new CouldNotConnectToRuntime(_host, _port);
```
jakhog, line 79: This is only for one specific exception, we should check only that or just let the original exception through.

Source/Services/Services.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="Grpc" Version="$(GrpcVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsVersion)" />
     <PackageReference Include="System.Reactive" Version="$(RxVersion)" />
+    <PackageReference Include="Contrib.Grpc.Core.M1" Version="$(GrpcVersion)" />
```
jakhog, line 15: I need to test this on my intel Mac.

Source/Services/IPerformMethodCalls.cs
```diff
@@ -25,6 +25,19 @@ public interface IPerformMethodCalls
...
+    /// <summary>
+    /// Performs the provided server streaming method call.
+    /// </summary>
+    /// <typeparam name="TClientMessage">Type of the <see cref="IMessage">messages</see> that is sent from the client to the server.</typeparam>
+    /// <typeparam name="TServerMessage">Type of the <see cref="IMessage">messages</see> that is sent from the server to the client.</typeparam>
+    /// <param name="method">The <see cref="ICanCallAServerStreamingMethod{TClientMessage, TServerMessage}"/> method to call.</param>
+    /// <param name="request">The <see cref="IMessage"/> to send to the server.</param>
+    /// <param name="token">A <see cref="CancellationToken"/> used to cancel the call.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    IServerStreamingMethodHandler<TServerMessage> Call<TClientMessage, TServerMessage>(ICanCallAServerStreamingMethod<TClientMessage, TServerMessage> method, TClientMessage request, CancellationToken token)
```
jakhog: Would be cool if return an IAsyncEnumerable here :)

jakhog (changes requested):
Lets go for async enumerables - they are nice

Source/Services/Services.csproj
jakhog: It works, they are put in the separate directories of the build output under `runtimes/<platform>/native` so they don't collide with each other.

jakhog (approved):

PR #120, opened 2022-01-24
Projection store

woksin:
- Changes API of `ProjectionStore.Get` to return only the TProjection
- Split out creation of projection store requests into its own thing
- Fix a bug in `ProjectionsToSDKConverter` so that it captures the execution returned when the current state type is invalid. 
- Fix a bug in `JsonSerializerEceptionCatcher` where it did not set the `ErrorContext.Handled` to `true` to make the `JsonSerializer`  not throw the Exception
- Add a Log static class for efficient logging for the `ProjectionStore`
- Remove `NoProjectionAssociatedWithType` Exception class as it is not needed anymore
- In `ProjectionStore.Get` throw `WrongKeyOnProjectionCurrentState` if the `Key` of a retrieved projection state has an unexpected Key


Source/Projections/Store/ProjectionStoreRequestCreator.cs
```diff
@@ -0,0 +1,45 @@
...
+
+/// <summary>
+/// Represents an implementation of <see cref="ICreateProjectionStoreRequest"/>.
+/// </summary>
+public class ProjectionStoreRequestCreator : ICreateProjectionStoreRequest
+{
+    readonly IResolveCallContext _callContextResolver;
+
+    /// <summary>
+    ///
```
jakhog: Missing docs

Source/Projections/Store/Converters/ProjectionsToSDKConverter.cs
```diff
@@ -29,38 +29,50 @@ public bool TryConvert<TProjection>(IEnumerable<PbCurrentState> source, out IEnu
...
-            PbCurrentStateType.CreatedFromInitialState => CurrentStateType.CreatedFromInitialState,
-            PbCurrentStateType.Persisted => CurrentStateType.Persisted,
-            _ => throw new InvalidCurrentStateType(type)
+            PbCurrentStateType.CreatedFromInitialState => (CurrentStateType.CreatedFromInitialState, default),
+            PbCurrentStateType.Persisted => (CurrentStateType.Persisted, default),
+            _ => ((CurrentStateType)pbType, new InvalidCurrentStateType(pbType))
         };
-}
\ No newline at end of file
+        return error == default;
```
jakhog: I think this would be easier to read with a switch case that sets the type or error, then returns inside the cases.

Samples/Tutorials/Projections/Program.cs
```diff
@@ -45,7 +45,7 @@
 var chef = await client.Projections
     .ForTenant(TenantId.Development)
     .Get<Chef>("Mrs. Tex Mex").ConfigureAwait(false);
-Console.WriteLine($"{chef.Key} has prepared {string.Join(", ", chef.State.Dishes)}");
+Console.WriteLine($"Mrs. Tex Mex has prepared {string.Join(", ", chef.Dishes)}");
```
jakhog: Use `chef.Name`.

jakhog (changes requested):

PR #128, opened 2022-02-06
Projections ioc

woksin:
Builds on #126. Adds IMongoDatabase and IMongoCollection<> for projection read models to the IoC container

### DIFF
--- a/Source/Projections/Builder/UnregisteredProjections.cs
+++ b/Source/Projections/Builder/UnregisteredProjections.cs
@@ -115,7 +115,7 @@ public class UnregisteredProjections : UniqueBindings<ProjectionModelId, IProjec
                         new object[]
                         {
                             MongoDBCopyCollectionName.GetFrom(readModelType).Value,
-                            new MongoCollectionSettings()
+                            null
                         }));
             }
         }
--- a/Source/Projections/Builder/UnregisteredProjections.cs
+++ b/Source/Projections/Builder/UnregisteredProjections.cs
@@ -17,6 +17,7 @@ using Dolittle.SDK.Projections.Store.Converters;
 using Dolittle.SDK.Tenancy;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


Source/Projections/Builder/UnregisteredProjections.cs
```diff
@@ -60,6 +64,16 @@ public UnregisteredProjections(IUniqueBindings<ProjectionModelId, IProjection> p
         }
     }
 
+    void RegisterProjectionsConventions()
+    {
+        var conventions = new ConventionPack();
+        conventions.AddClassMapConvention("Ignore extra projection properties", _ => _.SetIgnoreExtraElements(true));
+        ConventionRegistry.Register(
```
jakhog, line 71: Could this fail if you create two dolittle clients manually?

Source/Projections/Builder/UnregisteredProjections.cs
```diff
@@ -82,15 +96,28 @@ void AddToContainer(TenantId tenantId, IServiceCollection serviceCollection)
...
+            if (projection.Copies.MongoDB.ShouldCopy)
+            {
+                serviceCollection.AddSingleton(
+                    typeof(IMongoCollection<>).MakeGenericType(readModelType),
+                    serviceProvider => GetCollectionMethodForReadModel(readModelType).Invoke(
+                        serviceProvider.GetRequiredService<IMongoDatabase>(),
+                        new object[]
+                        {
+                            MongoDBCopyCollectionName.GetFrom(readModelType).Value,
+                            new MongoCollectionSettings()
```
jakhog: I think we should pass in `null` here so that the client uses the IMongoDatabase settings.
```suggestion
                            null
```

jakhog (changes requested):

Source/Projections/Builder/UnregisteredProjections.cs
jakhog: Nope, it just merges all the packs together. So it should be fine.

jakhog (approved):

PR #129, opened 2022-02-06
Conversions from classmap

woksin:
Based on #128 

Implements the conversions resolver from BSON class maps and fixes some specs and other implementations


Source/Projections/Builder/Copies/MongoDB/ConversionsFromBsonClassMapAdder.cs
```diff
@@ -0,0 +1,110 @@
+using System;
```
jakhog: The copyright header is missing, there are some other files with the same warning.

Source/Projections/Builder/Copies/MongoDB/ConversionsFromBsonClassMapAdder.cs
```diff
@@ -0,0 +1,110 @@
...
+                conversion = Conversion.DateAsString;
+                return true;
+            case BsonType.Int64:
+                conversion = Conversion.DateAsInt64;
+                return true;
+            default:
+                return false;
+        }
+    }
+    static bool IsComplexType(Type nominalType, IBsonSerializer serializer) => typeof(BsonClassMapSerializer<>).MakeGenericType(nominalType).IsInstanceOfType(serializer);
```
jakhog: Not sure if this is a problem, but maybe it is better to check if the `serializer` is an instance of `BsonClassMapSerializer<>` without the generic type specified? I think you can do something with `nominalType.GetGenericType()`. Just thinking about inheritance and serialisers for base classes.

Source/Projections/Builder/Copies/MongoDB/ConversionsFromBsonClassMapBuilder.cs
```diff
@@ -11,8 +12,19 @@ namespace Dolittle.SDK.Projections.Builder.Copies.MongoDB;
...
+    /// </summary>
+    /// <param name="conversionsFromBsonClassMap">The <see cref="IAddConversionsFromBsonClassMap" />.</param>
+    public ConversionsFromBsonClassMapBuilder(IAddConversionsFromBsonClassMap conversionsFromBsonClassMap)
+    {
+        _conversionsFromBsonClassMap = conversionsFromBsonClassMap;
+    }
+    
     /// <inheritdoc />
-    public bool TryBuildFrom<TReadModel>(IClientBuildResults buildResults, IPropertyConversions conversions)
+    public void BuildFrom<TReadModel>(IClientBuildResults buildResults, IPropertyConversions conversions)
```
jakhog, line 27: Not really an ask for a change, but just wondering in terms of performance. In this case, we could get away with the method taking in a `Type` as a parameter. But when doing it with generics like this, doesn't this mean that there will be one method generated for every type behind the scenes? If so, this might not be a best practice to keep using?

Source/Projections/Projections.csproj
```diff
@@ -19,6 +19,7 @@
         <ProjectReference Include="../Events/Events.csproj" />
         <ProjectReference Include="../Events.Processing/Events.Processing.csproj" />
         <ProjectReference Include="../Protobuf/Protobuf.csproj" />
+        <ProjectReference Include="..\Resources\Resources.csproj" />
```
jakhog: Just to keep style consistent.
```suggestion
        <ProjectReference Include="../Resources/Resources.csproj" />
```

Source/Resources/MongoDB/DolittleMongoConventions.cs
```diff
@@ -0,0 +1,40 @@
...
+    
+    /// <summary>
+    /// Ensured that the default Mongo Conventions are registered;
+    /// </summary>
+    public static void EnsureConventionsAreRegistered()
+    {
+        if (_isRegistered)
+        {
+            return;
+        }
```
jakhog, line 24: This operation shouldn't really take that long. So I think this early checking isn't all that necessary here.
```suggestion
```

Source/Resources/MongoDB/DolittleMongoConventions.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                return;
+            }
+            _isRegistered = true;
+            var pack = new ConventionPack();
+            pack.AddClassMapConvention($"Ignore Extras", _ => _.SetIgnoreExtraElements(true));
+            ConventionRegistry.Register(
+                "pack",
+                pack,
+                _ => true);
```
jakhog: We can be a little more descriptive:
```suggestion
            var pack = new ConventionPack();
            pack.AddClassMapConvention("Ignore extra elements", _ => _.SetIgnoreExtraElements(true));
            ConventionRegistry.Register("Dolittle Conventions", pack,  _ => true);
```

Specifications/Projections/Builder/Copies/MongoDB/for_ConversionsFromBsonClassMapAdder/when_adding_from_class_map/from_lookup_for_read_model_type/with_pascal_case_id_field.cs
```diff
@@ -0,0 +1,24 @@
...
+public class with_pascal_case_id_field : given.all_dependencies
+{
+    public class read_model
+    {
+        public string Id;
+        public string Field;
+        public string Property { get; }
+    }
+
+    Because of = () => adder.AddFrom(BsonClassMap.LookupClassMap(typeof(read_model)), build_results, conversions.Object);
```
jakhog, line 19: It would be nice if these specs passed in a mocked ClassMap, but we can change it later.

jakhog (changes requested):
Just some small things

Source/Projections/Builder/Copies/MongoDB/ConversionsFromBsonClassMapBuilder.cs
woksin: It seems like there is some overhead https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/generics-in-the-run-time :)

Source/Resources/MongoDB/DolittleMongoConventions.cs
woksin: I would maybe prefer to have it in, so that it follows the usual pattern of check => lock => check

Specifications/Projections/Builder/Copies/MongoDB/for_ConversionsFromBsonClassMapAdder/when_adding_from_class_map/from_lookup_for_read_model_type/with_pascal_case_id_field.cs
woksin: I was thinking of having that as additional specs to these ones with the types maybe?

Source/Projections/Builder/Copies/MongoDB/ConversionsFromBsonClassMapBuilder.cs
woksin: Let's not overdo it with Generics, but use it where it makes sense

jakhog (approved):

PR #130, opened 2022-02-09
Fetch resources when connecting, and expose a Task to wait for the client to be Connected.

jakhog:
## Summary

The Dolittle Client now fetches resources while establishing the initial connection so that we could make the resources interfaces synchronous, simplifying the usage and allowing us to more simply bind them in the DI container. The `.Connected` property on the client has been changed to a `Task` so you can await the connection asynchronously. The old boolean property has been moved to `.IsConnected`. 

### Added

- A new property `IDolittleClient.Connected` that returns a `Task` that is resolved when the client is successfully connected to a Runtime.

### Changed

- The `IMongoDBResource.GetDatabase()` returns an `IMongoDatabase` instead of a `Task<IMongoDatabase>` since the configuration is retrieved while connecting to the Runtime.
- The `IDolittleClient.Connected` boolean property has been renamed to `.IsConnected`.
- The `[MongoDBConvertTo(...)]` attribute was renamed to `[ConvertToMongoDB(...)]` as it was intended in the last release.

### Fixed

- Calling `IServiceProvider.GetDolittleClient()` could throw an exception if called while the client was in the process of connecting.


Source/Resources/Internal/ResourcesFetcher.cs
```diff
@@ -0,0 +1,50 @@
...
+    public async Task<IResourcesBuilder> FetchResourcesFor(IEnumerable<Tenant> tenants, CancellationToken cancellationToken = default)
+    {
+        var resources = new Dictionary<TenantId, IResources>();
+        
+        foreach (var tenant in tenants)
+        {
+            resources.Add(
+                tenant.Id,
+                new Resources(
+                    await _mongoDB.CreateFor(tenant.Id, cancellationToken).ConfigureAwait(false)
```
woksin, line 45: I'm thinking about a possible optimization here, instead of awaiting one by one we could do a Task.WhenAll(...) instead

woksin (approved):
I just have a comment about an optimization, but this is fine to merge either way

Source/Resources/Internal/ResourcesFetcher.cs
jakhog: Good point, done!

PR #132, opened 2022-02-10
Implement key selectors

woksin:
## Summary

Adds two new event key selectors to projections, `StaticKey` and `KeyFromEventOccurred`

### Added

- `[StaticKey]` event key selector attribute for projection On-methods that sets a constant, static, key as the key of the read model
- `[KeyFromEventOccurred]` event key selector for projection On-methods that uses the event occurred metadata as the key for the projection read models formatted as the string given to the attribute. We currently support these formats:
    - yyyy-MM-dd
    - yyyy-MM
    - yyyy
    - HH:mm:ss
    - hh:mm:ss
    - HH:mm
    - hh:mm
    - HH
    - hh
    - yyyy-MM-dd HH:mm:ss
    - And the above in different orderings


woksin:
TODO: Need to use the released contracts when that's ready

woksin:
This should now be ready for review

jakhog (approved):

PR #143, opened 2022-03-23
Small fixes in event handler convention builder

woksin:
### Fixed

- A bug where if there were errors while adding decorated event handler methods it would not try to add convention event handler methods


jakhog (approved):
:shipit:

PR #145, opened 2022-06-20
Tracing

woksin:
## Summary

Sets up tracing by default using Open Telemetry and exporting with OTLP. The tracing configuration is configurable through code using the `.UseDolittle()` API. By default the OTLP endpoint is set with the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. If that is not set it uses `http://localhost:4317`

### Added

- Automatic setup of tracing though OTLP by convention feeding traces and logs to endpoint set with the `OTEL_EXPORTER_OTLP_ENDPOINT` environment.
- Tracing on multiple important actions in the SDK like committing and handling events


Source/Execution/ExecutionContext.cs
```diff
@@ -25,4 +26,5 @@ public record ExecutionContext(
     Environment Environment,
     CorrelationId CorrelationId,
     Claims Claims,
-    CultureInfo Culture);
+    CultureInfo Culture,
+    ActivitySpanId? SpanId);
```
woksin:
```suggestion
    ActivitySpanId? SpanId = null);
```

woksin: This would entail a breaking API change if the last parameter is not optional

Source/SDK/Handshake/HandshakeResult.cs
```diff
@@ -10,4 +12,4 @@ namespace Dolittle.SDK.Handshake;
 /// </summary>
 /// <param name="MicroserviceId">The <see cref="MicroserviceId"/>.</param>
 /// <param name="Environment">The current <see cref="Environment"/>.</param>
-public record HandshakeResult(MicroserviceId MicroserviceId, Environment Environment);
+public record HandshakeResult(MicroserviceId MicroserviceId, Environment Environment, Uri? OTLPEndpoint);
```
woksin: Same as above
```suggestion
public record HandshakeResult(MicroserviceId MicroserviceId, Environment Environment, Uri? OTLPEndpoint = null);
```

PR #148, opened 2022-06-29
Fix resolution of singleton tenant scoped services 

woksin:
## Summary

A problem with our tenant scoped service providers caused crashes when resolving tenant scoped singleton services with dependencies not present in the derived tenant scoped service provider directly.

This turned out to be a problem deeply nested into how we try to integrate with Autofac using the capabilities of the standard Microsoft dependency injection framework. It does not allow us to fully take advantage of it. Therefore we're adding two extension packages for our DependencyInversion system, one for setting up the host to use Autofac and one for Lamar. We're also giving the developer the freedom to use their own dependency injection framework of choice by providing a method on `IHostBuilder` called  `UseDolittleTenantContainerCreator` that can be used to setup the system that creates the tenant scoped child containers from the root `IServiceProvider`, though we'd recommend settling for one of the extensions that we provide.

### Changed
- `IAggregateOf<>` registered as `transient` instead of `singleton` services 
- The Dolittle Client will explicitly throw an exception when connecting if there are services configured as Singleton that are not provided an instance due to limitations of the Microsoft dependency injection framework not having support for child containers. For that reason we have provided at the moment two extensions to DependencyInversion, Lamar and Autofac

### Added
- `UseDolittleTenantContainerCreator` method on IHostBuilder for setting up your own dependency injection framework with the tenant scoped provider setup if you don't want to use Autofac or Lamar
- Dolittle.SDK.Extensions.DependencyInversion.Autofac: Sets up Autofac as a service provider on the Host and enables the Dolittle Client to utilise it fully for per-tenant child containers
- Dolittle.SDK.Extensions.DependencyInversion.Autofac: Sets up Autofac as a service provider on the Host and enables the Dolittle Client to utilise it fully for per-tenant child containers
- Sample that shows how to set up DependencyInversion in Dolittle Client using Autofac, Lamar or neither

### DIFF
--- a/Samples/DependencyInjection/IGlobalTransientService.cs
+++ b/Samples/DependencyInjection/IGlobalTransientService.cs
@@ -1,4 +1,5 @@
 using System;
+using Microsoft.Extensions.Logging;
 
 public interface IGlobalTransientService : IDisposable
 {
@@ -6,4 +7,5 @@ public interface IGlobalTransientService : IDisposable
 
 public class GlobalTransientService : Service, IGlobalTransientService
 {
+    public GlobalTransientService(ILogger<TenantScopedTransientService> logger) {}
 }
--- a/Samples/DependencyInjection/ITenantScopedTransientService.cs
+++ b/Samples/DependencyInjection/ITenantScopedTransientService.cs
@@ -1,5 +1,6 @@
 using System;
 using Dolittle.SDK.Tenancy;
+using Microsoft.Extensions.Logging;
 
 public interface ITenantScopedTransientService : IDisposable
 {
@@ -7,7 +8,7 @@ public interface ITenantScopedTransientService : IDisposable
 
 public class TenantScopedTransientService : TenantScopedService, ITenantScopedTransientService
 {
-    public TenantScopedTransientService(TenantId tenant) : base(tenant)
+    public TenantScopedTransientService(TenantId tenant, ILogger<TenantScopedTransientService> logger) : base(tenant)
     {
     }
 }


woksin:
I'm a bit unsure why we did not add the registration source to the container when we did this, to me it seemed obvious that this is the solution. Though I'm not 100% certain. I need a couple more eyes and minds on this :)

Source/Aggregates/Builders/UnregisteredAggregateRoots.cs
```diff
@@ -39,7 +39,7 @@ void AddToContainer(TenantId tenantId, IServiceCollection serviceCollection)
     {
         foreach (var type in Types)
         {
-            serviceCollection.AddSingleton(
+            serviceCollection.AddTransient(
```
jakhog: This seems like a notable fix for the changelog?

Source/DependencyInversion/CannotRegisterSingletonDependenciesOnTenantScopedContainer.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// Exception that gets thrown when trying to register a singleton on a tenant scoped container.
+/// </summary>
+public class CannotRegisterSingletonDependenciesOnTenantScopedContainer : Exception
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CannotRegisterSingletonDependenciesOnTenantScopedContainer"/> class.
+    /// </summary>
+    public CannotRegisterSingletonDependenciesOnTenantScopedContainer()
+        : base("Currently registering singleton services that are not instances does not work on tenant scoped containers")
```
jakhog: A little more compact...
```suggestion
        : base("Singleton services that are not registered as a value, is not supported in per-tenant containers")
```

Source/DependencyInversion/TenantScopedProviders.cs
```diff
@@ -13,15 +14,15 @@ namespace Dolittle.SDK.DependencyInversion;
...
 
     /// <summary>
     /// Initializes a new instance of the <see cref="TenantScopedProviders"/> class.
     /// </summary>
     /// <param name="tenantScopedServiceProviders">The <see cref="IDictionary{TKey,TValue}"/> of <see cref="IServiceProvider"/> per <see cref="TenantId" />.</param>
-    public TenantScopedProviders(IDictionary<TenantId, IServiceProvider> tenantScopedServiceProviders)
+    public TenantScopedProviders(IDictionary<TenantId, (AutofacServiceProvider, IDisposable)> tenantScopedServiceProviders)
     {
-        _serviceProviders = new ReadOnlyDictionary<TenantId, IServiceProvider>(tenantScopedServiceProviders);
+        _serviceProviders = new ReadOnlyDictionary<TenantId, (AutofacServiceProvider, IDisposable)>(tenantScopedServiceProviders);
```
jakhog: I'm not super excited by the actual implementations leaking out. I would be more excited if the loop down below here just did a foreach, and maybe something like `if (provider is Disposable ...)`. I also have to admit that I really don't like the anonymous tuple `.Item1` and `.Item2` things.

Source/DependencyInversion/TenantScopedProvidersBuilder.cs
```diff
@@ -47,16 +47,20 @@ IServiceProvider CreateTenantContainer(TenantId tenant, IServiceProvider rootSer
...
         var container = containerBuilder.Build();
-
         var rootScope = container.BeginLifetimeScope(builder =>
         {
             builder.RegisterInstance(new ServiceScopeFactory(rootServiceProvider.GetRequiredService<IServiceScopeFactory>(), container)).As<IServiceScopeFactory>();
             builder.RegisterSource(new UnknownServiceOnTenantContainerRegistrationSource(rootServiceProvider));
         });
 
-        return new AutofacServiceProvider(rootScope);
+        return (new AutofacServiceProvider(rootScope), container);
```
jakhog: Related to the previous comment. It would be nice if we changed this to another implementation, maybe made a custom one so that when the service provider returned from this method disposes of the root scope when it is disposed. If it makes sense?

I might be rambling here, but it just feels like things are leaking around a little bit here...

Source/SDK/DolittleClientService.cs
```diff
@@ -13,7 +13,7 @@ namespace Dolittle.SDK;
 /// <summary>
 /// Represents an implementation of <see cref="IHostedService"/> for the <see cref="IDolittleClient"/>.
 /// </summary>
-public class DolittleClientService : IHostedService
+public class DolittleClientService : IHostedService, IDisposable
```
jakhog: Is this usually the way to get this thing done? Does the hosting thing dispose of these services automatically, or is it done by the container?

Another way would be to just dispose of the client when it is stopped?

jakhog (changes requested):
Not extremely important, but it feels to me like this disposable thing made a little bit of a mess. We can make it a bit cleaner :)

Also, I think the PR title and description is not really accurate in terms of what's been done now...

Source/Aggregates/Builders/UnregisteredAggregateRoots.cs
woksin: Yes, I was not quite finished yet 😁

Source/DependencyInversion/TenantScopedProviders.cs
woksin: I was a bit unsure how we wanted this, nice to get some feedback. I agree about the Item# part

Source/DependencyInversion/TenantScopedProvidersBuilder.cs
woksin: We can have IRootContainer or something that is disposable

Source/SDK/DolittleClientService.cs
woksin: We could do that as well, the IDisposable just ensures that the container disposes of it when the container is disposed to my knowledge

PR #149, opened 2022-07-01
Fix tenant containers 

woksin:
## Summary

A problem with our tenant scoped service providers caused crashes when resolving tenant scoped singleton services with dependencies not present in the derived tenant scoped service provider directly, pluss other problems. 

### Fixed
- The DependencyInversion internals to fully support tenant child containers for all kinds of IServiceProvider using Autofac as the dependency injection engine

### Changed
- `IAggregateOf<>` registered as `transient` instead of `singleton` services 

### Added
- `ICreateTenantContainers<>` interface for hooking in your own dependency injection library of choice as the tenant containers. Look at Samples/DependencyInjection for example for setting up Lamar
- Configuration methods for setting up the custom tenant container creator


Source/DependencyInversion/UnknownServiceOnTenantContainerRegistrationSource.cs
```diff
@@ -17,42 +18,94 @@ namespace Dolittle.SDK.DependencyInversion;
 /// </summary>
 public class UnknownServiceOnTenantContainerRegistrationSource : IRegistrationSource
 {
+    const string MetadataKey = "from-source";
```
jakhog: Just to reduce the chance of collision. We would really struggle to find this bug if someone made a registration with the same metadata.
```suggestion
    const string MetadataKey = "from-dolittle-unknown-registration-source";
```

jakhog (changes requested):

jakhog (approved):

PR #153, opened 2022-07-06
Fix problem with Artifact equality check

woksin:
## Summary

Fixes a problem with Artifacts caused event types with specified Alias to cause exception when being handled by an event handler

### Fixed

- EventType with explicit Alias should now work.


jakhog (approved):
:shipit:

PR #160, opened 2022-09-07
Aggregate events streaming

woksin:
## Summary

Adds a new method to the `EventStore` for fetching committed aggregate events filtered by event types. This allows us to change the rehydration of aggregates to be much more effective by just fetching the committed aggregate events for an aggregate that are relevant to the rehydration (meaning that there are an `On` method for that event type). This can have a significant impact on the performance of aggregates that have many events, but few state changes or are completely stateless. 

### Added

- `FetchForAggregate` that takes in a list event types used for filtering
- `FetchStreamForAggregate` fetches a stream of committed aggregate event batches

### Changed

- Rehydration logic of aggregate roots. It now only fetches the committed aggregate events that are relevant
- `AggregateRootVersion` on the `CommittedAggregateEvents` now represents the current aggregate root version of the aggregate root, not the version of the last committed aggregate event
- Minor version of Contracts meaning that this version of the SDK is only compatible with version `>= 8.5.0` of the Runtime

### Fixed

- Some users could experience exceptions when performing actions on aggregate roots that had lots of events or big events due to the protobuf messages being too big. This should be fixed now since the internals of fetching aggregate roots are now implemented using streaming and batching.

### DIFF
--- a/Source/Aggregates/AggregateRoot.cs
+++ b/Source/Aggregates/AggregateRoot.cs
@@ -50,10 +50,7 @@ public abstract class AggregateRoot
     /// </summary>
     public IEnumerable<AppliedEvent> AppliedEvents => _appliedEvents;
     
-    /// <summary>
-    /// Gets a value indicating whether this aggregate root is stateless or not.
-    /// </summary>
-    public bool IsStateless { get; }
+    bool IsStateless { get; }
 
     /// <summary>
     /// Apply the event to the <see cref="AggregateRoot" /> so that it will be committed to the <see cref="IEventStore" />
@@ -138,20 +135,14 @@ public abstract class AggregateRoot
     /// <summary>
     /// Rehydrates the aggregate root with the <see cref="CommittedAggregateEvents"/> for this aggregate.
     /// </summary>
-    /// <param name="aggregateRootId">The aggregate root id.</param>
     /// <param name="batches">The <see cref="IAsyncEnumerator{T}"/> batches of <see cref="CommittedAggregateEvents"/> to rehydrate with.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> used for cancelling the rehydration.</param>
     public async Task Rehydrate(IAsyncEnumerable<CommittedAggregateEvents> batches, CancellationToken cancellationToken)
     {
-        var batchesEnumerator = batches.WithCancellation(cancellationToken).GetAsyncEnumerator();
-        if (!await batchesEnumerator.MoveNextAsync())
+        var hasBatches = false;
+        await foreach (var batch in batches.WithCancellation(cancellationToken))
         {
-            throw new NoCommittedAggregateEventsBatches(AggregateRootId, EventSourceId);
-        }
-        
-        do
-        {
-            var batch = batchesEnumerator.Current;
+            hasBatches = true;
             ThrowIfEventWasAppliedToOtherEventSource(batch);
             ThrowIfEventWasAppliedByOtherAggregateRoot(batch);
             if (IsStateless)
@@ -165,7 +156,10 @@ public abstract class AggregateRoot
                 InvokeOnMethod(@event.Content);
             }
         }
-        while (await batchesEnumerator.MoveNextAsync());
+        if (!hasBatches)
+        {
+            throw new NoCommittedAggregateEventsBatches(AggregateRootId, EventSourceId);
+        }
         cancellationToken.ThrowIfCancellationRequested();
     }
 
--- a/Source/Aggregates/AggregateRoot.cs
+++ b/Source/Aggregates/AggregateRoot.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
+using System.Threading.Tasks;
 using Dolittle.SDK.Artifacts;
 using Dolittle.SDK.Events;
 using Dolittle.SDK.Events.Store;
@@ -16,6 +17,7 @@ namespace Dolittle.SDK.Aggregates;
 public class AggregateRoot
 {
     readonly List<AppliedEvent> _appliedEvents = new();
+    AggregateRootId? _aggregateRootId;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="AggregateRoot"/> class.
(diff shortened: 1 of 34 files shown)


woksin:
@jakhog Brought up a performance improvement that I@m gonna implement :)

Source/Aggregates/AggregateRoot.cs
```diff
@@ -15,7 +16,8 @@ namespace Dolittle.SDK.Aggregates;
 /// </summary>
 public class AggregateRoot
```
mhelleborg: Should this be abstract?

woksin: Maybe. Probably?

Source/Aggregates/AggregateRoot.cs
```diff
@@ -124,25 +136,47 @@ public void ApplyPublic(object @event, EventType eventType)
...
+    /// <param name="aggregateRootId">The aggregate root id.</param>
+    /// <param name="batches">The <see cref="IAsyncEnumerator{T}"/> batches of <see cref="CommittedAggregateEvents"/> to rehydrate with.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> used for cancelling the rehydration.</param>
+    public async Task Rehydrate(IAsyncEnumerable<CommittedAggregateEvents> batches, CancellationToken cancellationToken)
     {
-        ThrowIfEventWasAppliedToOtherEventSource(events);
-        ThrowIfEventWasAppliedByOtherAggregateRoot(events);
-
-        foreach (var @event in events)
+        var batchesEnumerator = batches.WithCancellation(cancellationToken).GetAsyncEnumerator();
```
jakhog: I think this enumerator needs to be disposed? Would it maybe be simpler with a `await foreach` loop?

Source/Aggregates/AggregateRoot.cs
```diff
@@ -42,6 +49,11 @@ public AggregateRoot(EventSourceId eventSourceId)
     /// Gets the <see cref="IEnumerable{T}" /> of applied events to commit.
     /// </summary>
     public IEnumerable<AppliedEvent> AppliedEvents => _appliedEvents;
+    
+    /// <summary>
+    /// Gets a value indicating whether this aggregate root is stateless or not.
+    /// </summary>
+    public bool IsStateless { get; }
```
jakhog: Does it need to be `public` - or could we make it a `private` variable? Do we use it outside of the `Rehydrate` method?

Source/Aggregates/AggregateRoot.cs
```diff
@@ -124,25 +136,47 @@ public void ApplyPublic(object @event, EventType eventType)
...
+            ThrowIfEventWasAppliedToOtherEventSource(batch);
+            ThrowIfEventWasAppliedByOtherAggregateRoot(batch);
+            if (IsStateless)
+            {
+                Version = batch.AggregateRootVersion;
+                break;
+            }
+            foreach (var @event in batch)
+            {
+                Version = @event.AggregateRootVersion + 1;
```
jakhog, line 155: I have to admit - I'm not sure here, are you sure there should be a `+1` before invoking the `On` method?

Source/Events/Store/Converters/AggregateEventToSDKConverter.cs
```diff
@@ -66,10 +65,9 @@ public bool TryConvert(Runtime.Events.Contracts.CommittedAggregateEvents source,
         {
             // We have to manually calculate and set the AggregateRootVersion for the events as the
             // CommittedAggregateEvents.AggregateRootVersion is set to the latest version.
```
jakhog: I believe this comment can now be deleted?
```suggestion
```

Specifications/Events/Store/Converters/for_AggregateEventToSDKConverter/given/committed_aggregate_events_and_a_converter.cs
```diff
@@ -31,7 +31,7 @@ public class committed_aggregate_events_and_a_converter : a_converter_and_a_prot
         event_source = "Rena Pearson";
         aggregate_root_id = "717915c1-bb88-4bec-b1c1-61451c5a6608";
 
-        aggregate_root_version = 186206759u;
+        aggregate_root_version = 0;
```
jakhog, line 34: Is there a particular reason why this version was changed?  Does it break something in this particular case?

Specifications/Aggregates/for_AggregateRoot/given/committed_events_and_two_aggregate_roots.cs
```diff
@@ -29,39 +30,54 @@ public abstract class committed_events_and_two_aggregate_roots : two_aggregate_r
...
+                    eventSource,
+                    aggregateRootId,
+                    i,
+                    i,
+                    @event,
+                    event_types.GetFor(@event.GetType()),
+                    false,
+                    executionContext)
+            });
+        }
```
jakhog: Just for completeness of testing, it could be nice if there was one batch with multiple events in it? Just to make sure we don't like only process the first one in the batch...

jakhog (changes requested):

Source/Aggregates/AggregateRoot.cs
woksin: It is not Disposable so no need for that :) The reason I had done it like that is for the check where it has no batches, but I guess that the cleanliness of the `await foreach` outweighs that :)

Source/Aggregates/AggregateRoot.cs
woksin: We can make it private

Source/Aggregates/AggregateRoot.cs
woksin: Yes, pretty sure :) Before it was a foreach loop for each event with `Version++` and `Version` starting on 0. This should be the same (first event has `AggregateRootVersion` = 0 etc.)

Specifications/Events/Store/Converters/for_AggregateEventToSDKConverter/given/committed_aggregate_events_and_a_converter.cs
woksin: I don't remember, I just remember it being broken

Specifications/Aggregates/for_AggregateRoot/given/committed_events_and_two_aggregate_roots.cs
woksin: Yes we should. The specs were just a little pain that's why this was a bit sloppy

Source/Aggregates/AggregateRoot.cs
jakhog: But it is `IAsyncDisposable` - don't they also need disposing?
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerator-1?view=net-6.0

woksin: Yeah, I noticed that as well in the Runtime, but when I checked the class that was returned here it did not seem to implement IDisposable. Weird. Either way I changed this to use await foreach, so it does not matter anymore :P

PR #162, opened 2022-10-19
Fix tenant scoped containers

woksin:
## Summary

Fixes for a problem with the tenant scoped containers where it could not resolve tenant scoped dependencies registered as callbacks.

### Fixed

- Tenant Scoped Containers


jakhog (approved):
:shipit:

PR #169, opened 2022-11-17
Actor hosted aggregates

mhelleborg:
## Summary

Rewrite of aggregate hosting on top of stateful Proto.Actor actors. Ensures single threaded access and removes the need to reload state on each perform (as long as the aggregate is loaded in memory). Will improve aggregate performance when multiple sequential (or parallel) calls to the same eventSourceId are executed.

### Added

- Dependencies on Proto.Cluster and Proto.OpenTelemetry

### Fixed

- Aggregates should no longer have any chance of AggregateRootConcurrencyConflicts within the same service


TomasEkeli:
merging this to get a pre-release built that we can play with

PR #193, opened 2022-12-21
Add automatic registration of PerTenant-Scoped services

woksin:
## Summary

Make it possible to automatically discover and register tenant-scoped services. This is not meant to be a replacement for the 
`WithTenantServices` method on the Dolittle configuration callback as that will still need to be used for services that requires more complicated registration, but the new attribute-based registration should suffice for most situations.
### Added

- `PerTenant` attribute that can be used to mark classes as PerTenant-scoped services. The `PerTenant` attribute has a couple of options in terms of `ServiceLifetime` (Transient by default) and whether the service should be registered as itself (off by default)

### DIFF
--- a/Source/DependencyInversion/TenantScopedProvidersBuilder.cs
+++ b/Source/DependencyInversion/TenantScopedProvidersBuilder.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Reflection;
+using BaselineTypeDiscovery;
 using Dolittle.SDK.Tenancy;
 using Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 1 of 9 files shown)


Source/DependencyInversion/TenantScopedProvidersBuilder.cs
```diff
@@ -64,4 +67,57 @@ IServiceProvider CreateTenantContainer(TenantId tenant)
...
+
+    static IEnumerable<Type> GetExportedTypes(Assembly assembly)
+    {
+        try
+        {
+            return assembly.ExportedTypes;
+        }
+        catch
+        {
+            return Enumerable.Empty<Type>();
```
mhelleborg, line 104: Log on failure?

woksin: We have pretty much this same code somewhere else as well. When it fails it's basically because the assembly could not be loaded, which probably in 99% of the cases is not interesting.

woksin: If we would have logging on these I would rather have the assembly scanning be cached or centarlized in some way and logged as Trace or something

woksin:
@mhelleborg Is this fine to release?

mhelleborg:
> @mhelleborg Is this fine to release?

Go for it, if it has been properly tested :)

PR #195, opened 2022-12-26
20.2.0 shire

woksin:
## Summary

Adds support for testing of Aggregate Roots in the Dolittle.SDK.Testing package (need to be referenced directly in test projects). Also adds a couple more quality-of-life tools to the SDK like MongoDB BSON serializer for ConceptAs types.

### Added

- Dolittle.SDK.Testing package with tools for testing Aggregate Roots
- MongoDB Bson serializer for ConceptAs types
- `WithMongoDatabaseSettings` method on Dolittle configuration builder for configuring `MongoDatabaseSettings`

### Fixed
- A bug with `PerTenant`-attribute where open-generic services would not be registered correctly


Source/Concepts/ConceptMap.cs
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) Dolittle. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Reflection;
+using Dolittle.SDK.Concepts;
```
mhelleborg, line 7: Cleanup usings?

woksin: I can do a usings cleanup of the whole solution and push it to master after this is merged

Source/Testing/Aggregates/AggregateRootAssertion.cs
```diff
@@ -0,0 +1,114 @@
...
+using System;
+using System.Linq;
+using Dolittle.SDK.Testing.Aggregates.Events;
+
+namespace Dolittle.SDK.Testing.Aggregates;
+
+/// <summary>
+/// Represents the assertions that can be performed on an <see cref="AggregateRoot"/>.
+/// </summary>
+public class AggregateRootAssertion
```
mhelleborg, line 14: This only asserts on type and count of the events. How do we verify that the details of an event is correct?

mhelleborg (approved):
This looks great, approved!

PR #218, opened 2023-05-31
Added Filtering of MongoDB Traces

KarolGrzesiak:
## Summary

This PR adds filtering of "useless" MongoDB traces, which are noisy, and don't provide much value.

### Added

- MongoClientSettingsExtensions

### Changed

- MongoDBResource

### DIFF
--- a/Source/Resources/MongoDB/MongoClientSettingsExtensions.cs
+++ b/Source/Resources/MongoDB/MongoClientSettingsExtensions.cs
@@ -21,7 +21,7 @@ static class MongoClientSettingsExtensions
     {
         HashSet<string> commandsToFilter = new(
             new[] { "ping", "isMaster", "buildInfo", "saslStart", "saslContinue" },
-            StringComparer.InvariantCultureIgnoreCase);
+            StringComparer.OrdinalIgnoreCase);
         builder.Subscribe(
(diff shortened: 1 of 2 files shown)


Source/Resources/MongoDB/MongoClientSettingsExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Initializes <see cref="DiagnosticsActivityEventSubscriber"/>, which creates traces based on MongoDB events, and adds it as subscriber to <see cref="ClusterBuilder"/>.
+    /// </summary>
+    /// <param name="builder">The MongoDB ClusterBuilder.</param>
+    public static void AddTelemetry(this ClusterBuilder builder)
+    {
+        HashSet<string> commandsToFilter = new(
+            new[] { "ping", "isMaster", "buildInfo", "saslStart", "saslContinue" },
+            StringComparer.InvariantCultureIgnoreCase);
```
mhelleborg: Should we do OrdinalIgnoreCase instead here? It's slightly faster since it's just comparing characters.

KarolGrzesiak: Agree, good point 👍🏻.

PR #220, opened 2023-06-08
Aggregate request deduplication

mhelleborg:
## Summary
Removed retries against aggregate actors. Added new test classes for Aggregates.

### Added
- Additional helpers / assertions for testing aggregate roots

### Fixed
- Removed potential retries on Aggregate.Perform, which could occur when response times are greater than the set retry timeout.


TomasEkeli (approved):
took a little while to understand the deduplication-thing: it's using proto middleware to make sure we don't actually send to two aggregates if the request is just a little slow

PR #221, opened 2023-06-12
OpenTelemetry SDK metrics

mhelleborg:
## Summary

Adds OpenTelemetry metrics to the SDK, enabled by default when OTEL is configured.

### Added
* AspNetCore metrics
* Dotnet runtime metrics
* Global metrics for Dolittle SDK for events committed, events processed/failed & aggregate events read / rehydrated.

### DIFF
--- a/Source/Diagnostics.OpenTelemetry/Diagnostics.OpenTelemetry.csproj
+++ b/Source/Diagnostics.OpenTelemetry/Diagnostics.OpenTelemetry.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <Import Project="../../default.props"/>
+    <Import Project="../../default.props" />
 
     <PropertyGroup>
         <AssemblyName>Dolittle.SDK.Diagnostics.OpenTelemetry</AssemblyName>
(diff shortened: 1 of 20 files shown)


Source/Diagnostics.OpenTelemetry/Diagnostics.OpenTelemetry.csproj
TomasEkeli, line 1: do we really want to depend on beta.1 and beta.3 -versions?

Source/Events.Handling/EventHandler.cs
TomasEkeli, line 1: why not use stopwatch.startNew and then elapsed?

Source/Events/Store/Internal/EventCommitter.cs
TomasEkeli, line 1: this class could benefit from some null-checking (not because of these changes)

Source/Diagnostics.OpenTelemetry/Diagnostics.OpenTelemetry.csproj
mhelleborg: Not really, but there does not exist any stable package for this. OTEL is moving fast, and not everything is nailed down yet :)

TomasEkeli:
a small change to use stop-watch instead of storing our own timestamps-as-longs. revert if there's some perf-reason that i'm not aware of.

also - do we want to depend on beta-versions for this?

otherwise: great!

Source/Events.Handling/EventHandler.cs
mhelleborg: I wanted the efficiencies from the .net 7-pattern of doing it, where we just capture the ticks. I agree that this is less readable tho, so I will use the older pattern

PR #222, opened 2023-06-12
Add_devcontainer

TomasEkeli:
## Summary

sets up devcontainer for developing on the dolittle sdk. development-environment -only. no changes to "real" code.

### Added

devcontainer-definition


TomasEkeli:
i've added a new label "no-release" for these kind of changes. not sure if the build-actions respect it.

PR #236, opened 2024-03-04
[DRAFT] Enable Using Return Types From Aggregates Methods

KarolGrzesiak:
## Summary

As a SDK user I would like to have a possibility of using Aggregates' methods in order to obtain information - this PR aims to start work on it. The design was done pretty quickly, and as a result of that is quite ugly, so just treat it as a starter, it probably can be done way better. 

### Added

- Return type from Perform Method in the AggregateRootOperations

### Removed

- Perform Method in the AggregateRootOperations, which takes Func<TAggregate, Task> as an input


mhelleborg:
Feature added in #237

